PR #478, opened 2014-12-30
dhcp4 user-class

kmroz:
Handle user-class option for dhcp in rfc3004 format as well as ad-hoc strings.

bnc#909307


mtomaschewski:
Hmm... unspecified value="2"? I'd say -->> take a closer look & discuss this again.

kmroz:
rc2 on the way... closing this.

PR #480, opened 2015-01-13
wireless: Proper dict entry for WIRELESS_AP is bssid (bsc#911315)

pwieczorkiewicz:


mtomaschewski:
No, this IMO does not work. It causes to never read the from client as the schema is using
access-point and thus wickedd does not send it to wpa-supplicant at all now...

pwieczorkiewicz:
Indeed it was not enough. I forgot about the schema again :-). According to the wpa_supplicant code for handling bssid:

static int wpa_config_parse_bssid(const struct parse_data *data,
                                  struct wpa_ssid *ssid, int line,
                                  const char *value)
{
        if (value[0] == '\0' || os_strcmp(value, "\"\"") == 0 ||
            os_strcmp(value, "any") == 0) {
                ssid->bssid_set = 0;
                wpa_printf(MSG_MSGDUMP, "BSSID any");
                return 0;
        }
        if (hwaddr_aton(value, ssid->bssid)) {
                wpa_printf(MSG_ERROR, "Line %d: Invalid BSSID '%s'.",
                           line, value);
                return -1;
        }
        ssid->bssid_set = 1;
        wpa_hexdump(MSG_MSGDUMP, "BSSID", ssid->bssid, ETH_ALEN);
        return 0;
}

the coming bssid value should be a string in the following format:
'\0', "\"\"", "any", "XX:XX:XX:XX:XX:XX"

Tested OK with the following values:
1. No WIRELESS_AP specified -> OK
2. WIRELESS_AP='' -> OK
3. WIRELESS_AP='any' -> OK
4. WIRELESS_AP='MY_AP_BSSID' -> OK
5. WIRELESS_AP='WRONG_BSSID' -> NOK

DELL:/etc/sysconfig/network # ifup wlp4s0
wicked: device wlp4s0 failed: operation timed out
wlp4s0          setup-in-progress

mtomaschewski:
This causes incompatible schema change without any true reason. I'd  keep access-point as it is.

pwieczorkiewicz:
Here is version w/o schema changing, but with ni_link_address_*() functions and preserving ARPHRD_ETHER type.

mtomaschewski:
works for me.

PR #481, opened 2015-01-14
extensions-hostname: return value to be based on outcome of hostname set

kmroz:
Since the syslog service is not installed by default on openSUSE systems,
the outcome of `rcsyslog reload` should not impact setting a hostname via
this Wicked extension.

### DIFF
--- a/extensions/hostname
+++ b/extensions/hostname
@@ -7,7 +7,6 @@ type=""
 family=""
 ifname=""
 
-set -e
 shopt -s nullglob
 cmd=$1; shift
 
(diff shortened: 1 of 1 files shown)


kmroz:
Marius,

Your patch was 99% there. I needed to remove the "set -e" at the top of the hostname extension as it was responsible for exiting the entire script when `rcsyslog reload` failed with non-zero return code.
Maybe you know some better way?

mtomaschewski:
"rcsyslog reload &>/dev/null  || :"

mtomaschewski:
I've overlooked the "set -e" initially -- basically all changes we need is IMO just the "|| :"

PR #482, opened 2015-01-16
xml: don't require parent's control node for cfg gener. (bsc#901517)

pwieczorkiewicz:
In case of XML config file, when parent node (bridge) does not have <control> node specified we do not generate it for the child device either. Otherwise it is cloned.
Children of bond device have control node by default generated (STARTMODE=hotplug).


mtomaschewski:
OK, the error seems  to be gone.

PR #483, opened 2015-01-19
dhcp4 user-class

kmroz:


kmroz:
We wanted to change the sysconfig variable formatting to adhere to standards... so new pull on the way.

PR #487, opened 2015-01-21
events: query and update device name on events  (bug#904061)

mtomaschewski:
A device name provided in a event may be obsolete in the meantime
(renamed, event backlog), thus explicitly query and update it on
newlink and udev event before we emit wicked device events.


pwieczorkiewicz:
looks and works ok for me

PR #491, opened 2015-01-28
compat: merge suse dhcp defaults and parse once

mtomaschewski:

PR #489, opened 2015-01-23
dhcp4 user class

kmroz:

PR #492, opened 2015-01-28
dhcp4: handle user-class option in config (bsc#909307)

mtomaschewski:
Handles parsing of rfc3004 formatted user-class
as well as non-standard but common string type.


mtomaschewski:
Reworked variant of [pull#489](https://github.com/openSUSE/wicked/pull/489).
See also [pull#491](https://github.com/openSUSE/wicked/pull/491).

kmroz:
Re-tested with pull#491. Works fine on my end and test suite reports no errors with pull#491+pull#492
(http://almond.arch.suse.de:8080/job/wicked-testing/118/changes)

PR #493, opened 2015-01-28
dhcp4: do not set/query mtu if set in the config (bsc#904323)

mtomaschewski:


okirch:
Just a quick comment, the part of the patch that handles print_doflags maps the "mtu" string to DO_NTP...

Olaf

mtomaschewski:
Thanks!
-       { "mtu",                DHCP4_DO_NTP            },
+       { "mtu",                DHCP4_DO_MTU            },

kmroz:
Works on my end also!

PR #504, opened 2015-02-11
support for timed fsm calls and ipv6 accept-dad sysctl

mtomaschewski:


mtomaschewski:
prerequires for bug 911562 and to avoid event backlog causing bug 905421

pwieczorkiewicz:
Looks good and works good for me. Test-suite also says OK.

PR #508, opened 2015-02-13
fsm: Add autocreated flag to the worker

pwieczorkiewicz:
Autocreated devices should be started by nanny,
whenver new policy for them arrives and device
already exists.

### DIFF
--- a/include/wicked/fsm.h
+++ b/include/wicked/fsm.h
@@ -137,6 +137,7 @@ struct ni_ifworker {
 				done		: 1,
 				kickstarted	: 1,
 				pending		: 1,
+				autocreated	: 1,
 				readonly	: 1;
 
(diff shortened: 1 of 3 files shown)


mtomaschewski:
Ahm... I'd say, dummy0 handling has to be improved / fixed inside wickedd...

pwieczorkiewicz:
Yes, this is just an easy to write workaround to this problem. Basically there are other option to solve that (device being either factory or unsoliticitaly created by the kernel) as well:
- adding some refreshes at ifup side
- adding some listner + handler + refresh at ifup side
- change nanny to always start up new policy when created (this is wrong IMO)

pwieczorkiewicz:
Not needed. Handled at wickedd some other way round.

PR #509, opened 2015-02-13
fsm: destroy action_table on when worker is done

pwieczorkiewicz:


mtomaschewski:
thanks, in testing now.

PR #510, opened 2015-02-13
Nanny: Handle rearming on important events in nanny only; Allow rearming from starting state of managed device

pwieczorkiewicz:


mtomaschewski:
this is something I'd defer to 0.6.17...

pwieczorkiewicz:
This will be re-worked and handled otherwise with new pull request.

PR #514, opened 2015-02-17
dummy: fix handling of automatically created dummy0

pwieczorkiewicz:
Initial dummy0[0] = if_nametoindex("dummy0") was not there
as it should be.


mtomaschewski:
Pawel, don't hurry -- it is not the only part we've missed in the merge of this unlucky commit,
there is still a race. Let's make it properly first.

PR #516, opened 2015-02-19
Client status output improvement; ifreload/ifup STARTMODE=manual proper handling

pwieczorkiewicz:


mtomaschewski:
Later, bug 918662 first.

mtomaschewski:
Needs  a rework due to  conflicts and we IMO should use a ni_ifworker_array_t ifmarked arrays
and not additional string arrays with names here.

pwieczorkiewicz:
Yes, rework is definitely needed.
We should use ni_ifworker_array_t ifmarked arrays as workers container for ifup/ifreload and ifdown procedures, but string arrays with ifnames are also needed to be sure (and properly process) interfaces lists user requested: e.g. eth0 eth1 all br1.

PR #518, opened 2015-02-26
logging: add pid option to syslog log-target

mtomaschewski:
Without options the pid is used, but syslog::stderr
resets it and syslog::stderr,pid was rejected...


pwieczorkiewicz:
Works for me. +1.

PR #522, opened 2015-03-09
service: ensure wicked does not run at the same time as networkd

pwieczorkiewicz:


oeffi:
we haven't installed networkd. Which rpm package does it belong to? I uninstalled NetworkManager, because we actually don't need this service. AFAIK neworkd is not part of OpenSuSE.
Uninstalling NetworkManager doesn't change anything at all.

mtomaschewski:
later.

PR #524, opened 2015-03-11
fsm: do not update the state of a worker on unexpected events

pwieczorkiewicz:


mtomaschewski:
Later, bug 918662 first.

mtomaschewski:
Most is included in pull #532, remaining things needs rework.

PR #525, opened 2015-03-13
sit,gre,ipip: drop state file instead of deleting object

pwieczorkiewicz:


mtomaschewski:
Yes, looks like a bug & correct fix -- a DELLINK ACK event will delete it along with netdev after all events client is waiting for are processed & send out. Apply to testing branch.

pwieczorkiewicz:
Applied to testing branch.

PR #526, opened 2015-03-13
sit,ipip: Unload module when ifdown on sit0 or tunl0

pwieczorkiewicz:
Since we are loading sit/gre/ipip modules on ifup, it _might_ make sense to unload it and therefore allow for graceful ifdown of sit0 or tunl0. Otherwise kernel does not let to remove them and ifdown (especially one with --delete aka --force device-delete) waits 30s to timeout.

### DIFF
--- a/dhcp4/main.c
+++ b/dhcp4/main.c
@@ -264,7 +264,7 @@ main(int argc, char **argv)
 	ni_srandom();
 
 	/* load af_packet module we need for capturing */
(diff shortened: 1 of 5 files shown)


pwieczorkiewicz:
After review both Reviewer and Author agreed that this PR sucks... closed.
Sit0 special case (kernel does not allow to delete it unless sit module is removed) should be handled somewhere else (kernel probably).

PR #530, opened 2015-03-23
Fsm: Introduce ni_fsm_ifworker_find() to looks for workers (bsc#918662)

pwieczorkiewicz:
Also additional fixes:
- do not destroy worker from nanny event handler, but only from fsm event handler
- assign w->ifindex on createDevice() call (calculate from object_path)


mtomaschewski:
This does not really makes it better yet.

pwieczorkiewicz:
Yes, this is not enough (just a step towards right direction), further improvements are needed.

mtomaschewski:
Obsoleted by pull #532.

PR #532, opened 2015-04-02
fsm: fixes to use only ready devices by name and event races

mtomaschewski:
Instead to add not yet ready devices to the normal fsm worker set, add them into a special
pending worker set to avoid confusions while resolving dependencies by name before or
while udev renames them and thus assign wrong config to devices (bsc#918662) and handle
related event race conditions (bsc#921218).

### DIFF
--- a/include/wicked/fsm.h
+++ b/include/wicked/fsm.h
@@ -223,6 +223,7 @@ struct ni_fsm_require {
 };
 
 struct ni_fsm {
+	ni_ifworker_array_t	pending;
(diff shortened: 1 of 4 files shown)


mtomaschewski:
Picked up relevant commits from pull#523 and pull#524. Obsoletes pull#530.

pwieczorkiewicz:
customer verified OK in bug 918662 comment 80, test-suite passed, remaining issues in further pull request

PR #535, opened 2015-04-14
fsm: optimization for ni_fsm_wait_tentative_addrs()

pwieczorkiewicz:
- refresh only particular worker when necessary
- wait for succeeded workers without a link, when link is required
- reduce the amount of printouts


mtomaschewski:
later

PR #542, opened 2015-04-29
spec: Add requirement to util-linux-systemd for wicked-service (bsc#911310)

pwieczorkiewicz:


mtomaschewski:
a pre-req, see bug report. on update, there are old sysconfig scripts which we execute while migration.
I'd even add a prereq to wicked package already, because our ifup wrapper scripts use logger too....

mtomaschewski:
note also, that this is just about an error message generated by old packages -- does not cause any harm. -> low prio.

pwieczorkiewicz:
Yes. Applied changes and corrections as suggested. Putting on hold for now.

mtomaschewski:
I'd say let's accept it -- it should not hurt, just ensures that the correct package providing logger is installed before wicked.

PR #543, opened 2015-04-30
spec: conditionally enable nanny use (boo#915025,bsc#916402)

pwieczorkiewicz:


mtomaschewski:
I didn't tested it, but it looks fine -> put it to testing branch and if OK, we can accept it.

PR #545, opened 2015-06-09
dhcp6: refresh ipv6 link on newprefix in auto mode (bsc#934067)

mtomaschewski:
When NEWPREFIX arrives, there definitely were an RA, but
when the RA contained a RetransTimer/ReachableTime of 0
(unspecified / no change to before), the kernel forgets to
send a NEWLINK event, even other things in the RA, as
the managed/other-config flags we wait for, changed.

### DIFF
--- a/dhcp6/device.c
+++ b/dhcp6/device.c
@@ -41,6 +41,7 @@
 #include "dhcp6/fsm.h"
 #include "appconfig.h"
 #include "util_priv.h"
+#include "netinfo_priv.h"
 #include "duid.h"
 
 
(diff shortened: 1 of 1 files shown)


mtomaschewski:
Olaf, does it workaround the issue you ran into?

okirch:
Hi Marius,

I'm travelling right now; hopefully I'll get an oppty to test it next week

Thanks!
Olaf

-- 
Nous sommes du soleil
We love when we play

pwieczorkiewicz:
Works for me.

PR #550, opened 2015-06-11
FSM: Improve handling of the DEVICE_DELETE event (hotplug oriented)

pwieczorkiewicz:


pwieczorkiewicz:
Commit 9051fcc fixes the race condition which occurs with the previous version if this pull.
The main idea behind it is:
- bind->method is executed in a call and must be present
- bind->service is only needed there for its name (bind->service->name)
- after the call given worker might be erased, thus bind structure might be purged.
- since we need only names of the method and service after the call, we can as well work with the copies.
- ni_ifworker_update_from_request() is only relevant for Addrconf calls (RequestLease, DropLease, ...).

PR #551, opened 2015-06-11
fsm: fix memleaks of w->device_api.config and binding[].config

pwieczorkiewicz:


mtomaschewski:
This is more complex than the  commit suggests -- we need much more than this to cleanup/fix it properly. I'd propose to close this pull request and then work on a new patch set doing it....

pwieczorkiewicz:
Yes, after discussing the problem behind this patch, we decided to take different approach here.

PR #553, opened 2015-06-12
wireless: send passwd when phase2 method isn't set (boo#927615)

pwieczorkiewicz:


bmwiedemann:
:+1: worked in my test on Factory

bmwiedemann:
maybe NI_WIRELESS_EAP_NONE would better be named NI_WIRELESS_EAP_ANY - could reduce future confusion.

PR #556, opened 2015-06-17
policy: fix to worker must be ready to satisfy match check

pwieczorkiewicz:
Reject worker (or child) in the <match> check only when:
- device has been created for a given worker,
  but it isn't ready.
- it's a config worker

This commit fixes ae67663af679414d2020437689d22cacefb5d92f.


mtomaschewski:
OK, pull#550 + pull#556 worked fine.

PR #557, opened 2015-06-17
fsm: use netif-reference shared flag as relation indicator in meta:require too

pwieczorkiewicz:
This is needed in order to catch and order sequence of FSM action steps properly for scenarios where a device has both master and lower device dependency. That it a bond0.21 vlan interface needs to wait for its lower bond0 interface to be at least in device-up state _and_ it has to wait till its master (e.g. lab21) interface is present to it can receive single second NEWLINK with both relationships set.


mtomaschewski:
OK, my trinity seems to work fine now -> testing branch.

mtomaschewski:
yes, there is no need to consider relation type when adding requirement.

PR #558, opened 2015-06-19
xml, dbus: hide less relevant logs under debug2 and  3 level

pwieczorkiewicz:


mtomaschewski:
just some debug output tweaks -> OK, but later.

PR #565, opened 2015-07-09
fsm: update worker's iftype on refresh

pwieczorkiewicz:


mtomaschewski:
what I see is: throw an error and fail.

pwieczorkiewicz:
In that pull firstly check if dev->link.type is set and valid:
 if not then:
   normalize it to UNKNOWN; 

then check if worker's iftype is different than that from dev->link.type
  if not then:
    do nothing;
  otherwise:
    if worker's iftype is UNKNOWN or not valid then:
      update it from dev->link.type;
    otherwise:
      fail /*since worker's iftype is normal, but differs from it's refreshed device linktype (some strange error)*/

mtomaschewski:
Please provide a concrete testcase showing this scenario, so we/I can  take a look what happens and whether this is the right place to handle it. a fail is IMO ok, but return NULL breaks e.g. event handling.

mtomaschewski:
later

PR #566, opened 2015-07-11
nanny: correct workers starting order and do not restart FSM on slaves/ports

pwieczorkiewicz:
This pull request has dependency to the #567 .

### DIFF
--- a/client/ifdown.c
+++ b/client/ifdown.c
@@ -243,6 +243,10 @@ usage:
 		return NI_WICKED_RC_ERROR;
 	}
 
+
+	ni_ifworker_array_flatten(&fsm->workers);
+	ni_ifworker_array_depth_sort(&fsm->workers);
+
 	/* Get workers that match given criteria */
 	nmarked = 0;
 	while (optind < argc) {
--- a/client/ifdown.c
+++ b/client/ifdown.c
@@ -57,6 +57,7 @@ ni_ifdown_stop_policy(const char *policy_name)
 	return TRUE;
 }
 
+#if 0
 static ni_bool_t
 ni_ifdown_stop_device(const char *device_name)
 {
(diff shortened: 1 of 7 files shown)


mtomaschewski:
It looks like symptomatic fixes that do not correct the root causes, so I'd  defer this and address the root causes first and consider this as a "candidate collection".

pwieczorkiewicz:
Yes. That's true. This patchset consists of:
- loopback detection (already discussed how to improve that bit)
- dynamically assign depth parameter on workers' structure changes and (re)sort fsm worker array when then. This gives us loopback worker always as a first worker in the array and ensures more "sane" worker order so fsm starts as expected. Not so important fix, but currently depth sometimes is assigned wrongly, and workers' order is quite random (which is not a bug though).
- do not apply policies in reveres order form ifup
- fix multiple FSM runs on slaves/ports in nanny.
- dont disable() devices on ifdown, since events take care of rearm 
- block events when hierarchy is being rebuilt (ensures the integrity, bit not a real problem at the moment)
- ni_check_state_req_data into structure cleanup
- rebuild hierarchy on DEVICE_READY before register at nanny

mtomaschewski:
The fix for multiple FSM runs on slaves/ports in nanny is IMO an important win, this was causing
really many useless loops. handling events in nanny after hierarchy rebuild  makes sense so it
gets the actual tree instead of the old one, so let's put to testing and apply.
I didn't found any  regressions in my tests of the previous version, cleaned up a bit it looks better.

mtomaschewski:
Unfortunately, testsuites fail in several complex places -- especially under nanny :-/

pwieczorkiewicz:
Yes. In all these places it is due to link/master relation missing from port/slave. With pull#567 which covers lack of this in nanny recheck() testsuite is green.
We have to fix the relation problem first then.

mtomaschewski:
later

PR #567, opened 2015-07-11
nanny: add masterdev|lowerdev dependency checks to recheck()

pwieczorkiewicz:
This pull adds extra checks to recheck() which prevents starting FSM prematurely when workers hierarchy dependences are not yet resolved. That helps with situations when workers are added in different order (ports before master; vlan before lowerdev); this is especially visible during hotplug scenarios.


mtomaschewski:
Hmm... IMO simptomatic changes -- it is not so much about matches policies per se, but about completing missed relation in the config before it arrives fsm. When these are incomplete, the policy will be incorrect as  side effect... So IMO also a  collection to pick commits from it as needed...

pwieczorkiewicz:
The above changes have more importance when dealing with hotplug cases. When master or lower condition is not met, no need to start FSM to find that out. Also if we start a slave of a bond that just has disappeared and it appears again during slave FSM cycle we may see problems.

pwieczorkiewicz:
Discussed to be dropped in favour of linkmaster change already accepted. It is better to handle such dependencies in FSM rather than at nanny's recheck().

PR #568, opened 2015-07-11
nanny: add min and max state requirements to match master and lower checks

pwieczorkiewicz:
This pull request has strong dependency to the #567 and soft to #566 .
Should not be applied without #567 (89766a7 and 1e27d24 are parts of #567 already).

This pull adds extra state checks into <master> and <lower> nodes of <match> in order to make recheck() more picky when to start given worker. This change is optional since FSM provides same (even more advanced) functionality, but thanks to this change nanny is able to delay starting FSM on devices which could wait until minimal requirements are met (e.g. vlan waits for lowerdev to be in device-up state, before it's created).


mtomaschewski:
Hmm... looks like a further set of symptomatic fixes, similar to  pull#566 and  pull#567 -- adds new check constructs without to fix existing...

pwieczorkiewicz:
Yes, this patch set is just an optional enhancement. It does not fix anything. Just optimizes a bit recheck(), which we may not want in the end...

pwieczorkiewicz:
Discussed to be dropped. No need for such extra handling. It's FSMs job.

PR #571, opened 2015-07-15
fsm: reference count worker config nodes to avoid memleaks

mtomaschewski:


pwieczorkiewicz:
LGTM. Test-suite is green.

PR #572, opened 2015-07-20
Improve nanny related calls: create/replace/update/delete policy

pwieczorkiewicz:
- add replacePolicy call and use it instead of create; it accepts multiple policies at once and auto restart given workers.
- remove old createPolicy call
- fix and improve updatePolicy (old update()) call
- fix and improve deletePolicy call

- lots of minor tweaks and cleanups


mtomaschewski:
we need something like this, but it needs very close look and IMO rework, later.

PR #573, opened 2015-07-20
fsm: ifworker status check functions tweaks

pwieczorkiewicz:


mtomaschewski:
tweaks -- again, really? please address root causes instead change these functions again and again

PR #579, opened 2015-08-10
Ifstatus: fix no devices output handling (bsc#939142)

pwieczorkiewicz:


mtomaschewski:
yes, this is about the case when none of the requested interfaces is there -> looks fine, please add to testing.

PR #583, opened 2015-08-28
compat: OVS_BRIDGE option parsing and generation (fate#318840)

kmroz:
Needed for ovsbr-conf script to determine which ovs bridge a port belongs to - during ifup/ifdown - allowing for add-port/del-port operations.

### DIFF
--- a/client/compat.c
+++ b/client/compat.c
@@ -1775,6 +1775,12 @@ __ni_compat_generate_ifcfg(xml_node_t *ifnode, const ni_compat_netdev_t *compat)
 		}
 	}
 
+	if (dev->ovs_bridge) {
+		xml_node_t *child;
+		child = xml_node_create(ifnode, "ovs-bridge");
+		xml_node_new_element("name", child, dev->ovs_bridge);
+	}
+
 	switch (dev->link.type) {
 	case NI_IFTYPE_ETHERNET:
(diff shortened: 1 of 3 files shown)


kmroz:
To have a look at the OVS side changes, please go to: https://build.suse.de/package/show/home:k_mroz:branches:SUSE:SLE-12-SP1:GA/openvswitch

There you will find:

1. ovsbr-conf script
2. Changes to the spec file to install the script under /etc/wicked/scripts/ovs
3. Change to openvswitch.service to start OVS before=network.service, allowing shutdown to happen after network service. This ensures connectivity for NFS, etc during system shutdown.

To see a live system with wicked:master + this pull, and OVS from the above mentioned branch, please a look at drnek.

mtomaschewski:
I think, we have to use an ovs extension allowing to have an own <ovs:port> and maybe also <ovs:bridge> config nodes, not pre-up/post-up scripts which the user have to configure additionally.
... I'll look into this tomorrow.

mtomaschewski:
We've tried to implement it via (extension) scripts, but this does  not work properly because of the invisible dependencies  and so on. Thus I rewrote the extension in C, and added already some dependency handling -- works much better IMO, see pull#https://github.com/openSUSE/wicked/pull/585.

PR #587, opened 2015-09-11
dhcp4: verify client-id if send back by the server

mtomaschewski:
RFC 6842 requires that servers MUST send back client-id as
sent by the client and that client MUST verify it, when it
receives it in server responses.
Some servers need a config option (e.g. echo-client-id on
recent ISC dhcp) to send client-id back.


pwieczorkiewicz:
LGTM.

PR #594, opened 2015-09-22
wireless: encode/decode ESSID as hex (bsc#941708)

mtomaschewski:


mtomaschewski:
ok, this is not sufficient ... :-/

PR #596, opened 2015-10-08
dhcp4: fix to request offer by default in --test (bsc#942278)

mtomaschewski:


mtomaschewski:
Regression caused by  commit 10e71b1 in [pull#288](https://github.com/openSUSE/wicked/pull/288):
  ``dhcp4: recover lease and correctly build requests (bnc#881475)``

PR #597, opened 2015-10-08
dhcp: moved most dhcp sources to the library

mtomaschewski:
Preparation permitting to call dhcp tester directly in a wicked test command.


mtomaschewski:
applied to testing to let the testsuite run.

mtomaschewski:
All testsuite runs was fine with this move (it does not introduce true changes yet).

PR #598, opened 2015-10-08
nanny: add device references to policy match (bsc#941611)

mtomaschewski:
Generate a match to all devices a config refers to in it's link node
(master device, ovs-bridge, ...) to ensure, that there is already
the complete configuration tree applied to nanny in order to not
start managing a device and fail on unresolvable requirements.


mtomaschewski:
No, this needs a change / fix.

mtomaschewski:
OK, this works fine for me now... We'll see what the testsuite results are.

mtomaschewski:
testsuite are fine with this.

PR #599, opened 2015-10-13
client: add wicked test dhcp[46] commands (bsc#942278)

mtomaschewski:


client/tester.c
```diff
@@ -0,0 +1,331 @@
+/*
+ *      wicked client tester commands
+ *
+ *      Copyright (C) 2014 SUSE LINUX Products GmbH, Nuernberg, Germany.
```
aaannz: Not that it's any of my business, but why do people insist on obsoleted copyrights? OpenQA suffers from this quite often too. I think when Wicked pictures will make second run for wicked we should have it all in order ;)

According to most recent email from legal it should be like
`Copyright (C) 2015 SUSE LLC`

PR #602, opened 2015-10-16
fsm: no link-up check on unrelated master (bsc#948423)

mtomaschewski:
When the lower device (bond0) of the one we want to ifup (vlan1 on
bond0), has an unrelated/inactive in this ifup run, but sufficiently
configured master (br0 on bond0), continue instead to follow and
enforce link-up/enslave requirements on this unrelated master.


pwieczorkiewicz:
LGTM. Tested OK.

PR #603, opened 2015-10-19
netconfig: do not refresh unrelated link details of address families

mtomaschewski:
dhcp6 does not need IPv4 related state, dhcp4 and auto4 do not need IPv6.


mtomaschewski:
testsuite run on dante were successful

pwieczorkiewicz:
Tested OK with the test-suite.

PR #604, opened 2015-10-20
service: fixed wicked client and nanny dependencies (bsc#950333)

mtomaschewski:
Fixed to want nanny in wicked client (rcnetwork) service, added explicit before
wickedd-nanny ordering to wickedd service and changed to use wants instead
of wantedby install dependencies which are active for enabled services only.


pwieczorkiewicz:
Tested OK.

PR #606, opened 2015-10-21
policy: fix policy file reading and policy objects references (boo#916035)

pwieczorkiewicz:


mtomaschewski:
IMO we can accept it as soon as the 2nd testsuite run was successful.

PR #610, opened 2015-11-18
nanny: skip recheck when no policy configured

mtomaschewski:


mtomaschewski:
names are verified non-empty & valid already before (!name just for completeness),
later code will catch it and jump out, so IMO there is no harm caused by this bug,
just some useless lookups.

pwieczorkiewicz:
LGTM. No need to recheck non-existing policy.

PR #618, opened 2015-12-16
fsm: do not segfault on device hierarchy loops (bnc#959356)

pwieczorkiewicz:
Since fsm check requirements are dynamic, we do not need to enforce any order of workers or policies being processed. Thus flatten functionality can be removed.

The fsm worker hierarchy loops should be checked earlier and on every hierarchy rebuild. That way we can always detect misconfiguration and fail affected workers.

### DIFF
--- a/client/ifreload.c
+++ b/client/ifreload.c
@@ -662,7 +662,6 @@ usage:
 	}
 
 	ni_fsm_pull_in_children(&up_marked);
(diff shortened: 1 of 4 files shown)


mtomaschewski:
OK, this removes the function causing the segfault, but unfortunately it also causes ifup to fail completely and not setup any network interface, also not completely unaffected branches of the tree and e.g. loopback.

PR #620, opened 2016-01-18
fsm: omit pending wait for hotplug devices (boo#945601)

pwieczorkiewicz:


mtomaschewski:
I don't know what it causes exactly without to investigate, but this does not address the referenced issue. IMO the issue is **not** about the fsm, but about (monitoring the progress and checking status while) waiting in ifup. ifup still waits the full timeout (60sec in the case bellow).

Example with 3 configured devices here:
- nic0 + lo using STARTMODE=nfsroot (_wait_ for them at start and skip
normal ifdown requests as an addition to STARTMODE=auto alias boot] and an
- eth0 using STARTMODE=hotplug (... an optional usb2eth device, which is usually not even
plugged in; do not wait for it when device does not exists/is not ready)

```
2016-01-29T12:02:15.900116+01:00 xanthos wicked[23208]: Executing: wicked --systemd --debug all --log-target syslog::stderr,pid ifup --timeout 60 all
[....]
2016-01-29T12:02:17.520339+01:00 xanthos wickedd-nanny[6480]: nic0: successfully transitioned from addrconf-up to network-up
2016-01-29T12:02:17.520596+01:00 xanthos wicked[23208]: received signal progressInfo; object_path=/org/opensuse/Network/Nanny/Interface/2; target_state=network-up, state_name=network-up
2016-01-29T12:02:17.520873+01:00 xanthos wickedd-nanny[6480]: waiting for 0 devices to become ready (0 explicitly requested)
2016-01-29T12:02:17.521153+01:00 xanthos wickedd-nanny[6480]: ni_nanny_recheck(eth0)
2016-01-29T12:02:17.521404+01:00 xanthos wickedd-nanny[6480]: eth0: no applicable policies
###
### we're DONE here in this case, because eth0 is not plugged in / there is nothing we could do.
###
2016-01-29T12:03:15.987217+01:00 xanthos wicked[23208]: Interface wait time reached
2016-01-29T12:03:15.997233+01:00 xanthos wicked[23208]: Exit with status: 0
```

mtomaschewski:
I'd say, we can reopen when we have a fix at the right place.

PR #624, opened 2016-02-03
schema: require bond slaves in device-setup (bsc#964877)

mtomaschewski:


mtomaschewski:
Even correct IMO, seems to make even worser :-/ Don't merge.

PR #625, opened 2016-02-22
fsm: do not use master name after reset (bsc#959356)

mtomaschewski:


pwieczorkiewicz:
LGTM.

mtomaschewski:
This bug has been introduced in commit adf8004 from pull #618 and effectively replaces one segfault with another one.

PR #638, opened 2016-04-13
auto6: receive IPv6 RA DNSSL forwarded by kernel

mtomaschewski:


mtomaschewski:
This is just some cleanup of the event receive part of the branch without further processing,
that is there is no lease written (to resolv.conf) yet.

PR #649, opened 2016-05-13
Initial implementation of PPPoE (fate#317976, bsc#865573, boo#913861)

pwieczorkiewicz:


mtomaschewski:
OK, this works so far in demand/persist mode using ifcfg created by yast2-network on sles11.

PR #660, opened 2016-07-21
dhcp4: initial support to request custom options (bsc#916948,bsc#988954)

mtomaschewski:

PR #672, opened 2016-10-10
dhcp: fixed to properly append dhcp option data

mtomaschewski:


mtomaschewski:
a segfault introduced by pull #660 :-(

PR #675, opened 2016-11-04
dhcp: support to define and request custom options (bsc#988954)

mtomaschewski:


mtomaschewski:
TODO: documentation how to do it

mtomaschewski:
This pull request needs a  fix (and documentation... work-in-progreess now).
The ipv6-prefix type parsing and formatting should be as defined in https://tools.ietf.org/html/rfc7227#section-5.3 instead to be just an shortcut for ipv6-address + prefix-len ...

mtomaschewski:
OK, adjusted, incl embedded-length attribute to encode { len, data } in string and opaque types + man pages.

Issue #681, opened 2016-11-23
wicked is slow to start up at boot


rombert:
Running on openSUSE Tumbleweed x86_64

    $ rpm -q wicked
    wicked-0.6.39-1.1.x86_64

Booting up is quite slow, and `systemd-analyze` points to wicked

    # systemd-analyze blame | head -n 3
         16.792s wicked.service
          2.081s logrotate.service
          1.163s apparmor.service

The journal does not say much, pasting below the output. There is a large gap between "Committed DHCPv4 lease" and "Notified neighbours about IP address" though:
```

# journalctl -b | grep wicked
Nov 23 22:52:16 mars systemd[1]: Starting wicked AutoIPv4 supplicant service...
Nov 23 22:52:16 mars systemd[1]: Starting wicked DHCPv4 supplicant service...
Nov 23 22:52:16 mars systemd[1]: Starting wicked DHCPv6 supplicant service...
Nov 23 22:52:16 mars audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 msg='unit=wickedd-dhcp6 comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Nov 23 22:52:16 mars systemd[1]: Started wicked DHCPv6 supplicant service.
Nov 23 22:52:16 mars audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 msg='unit=wickedd-dhcp4 comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Nov 23 22:52:16 mars systemd[1]: Started wicked DHCPv4 supplicant service.
Nov 23 22:52:16 mars audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 msg='unit=wickedd-auto4 comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=success'
Nov 23 22:52:16 mars systemd[1]: Started wicked AutoIPv4 supplicant service.
Nov 23 22:52:16 mars systemd[1]: Starting wicked network management service daemon...
Nov 23 22:52:16 mars systemd[1]: Started wicked network management service daemon.
[...]

PR #682, opened 2016-11-25
dhcp6: defer on prefix event if RA flags are unset

mtomaschewski:
When a RA prefix event arrives and the managed/other-config
RA flags are not set (dhcp not used on the network), send a
defer event to stop waiting for the lease in ifup earlier (issue #681)


mtomaschewski:
needs a closer look -> later

PR #699, opened 2017-03-28
bonding: properly send primary reselect as U8 (bsc#1027231)

mtomaschewski:


mtomaschewski:
Rainhard,
could you please verify also all the another PUT against the bond_policy[]
table in $kernel-source/drivers/net/bonding/bond_netlink.c? Thanks!

rma-x:
All instances of NLA_U* are matching between wicked and kernel, but I don't know how NLA_NESTED and NLA_BINARY should be handled in userland.

PR #709, opened 2017-06-13
WIP: Implement new ethtool offload options: rxvlan, txvlan, ntuple and rxh…

rma-x:
…ash.


mtomaschewski:
when the config contains multiple options, e.g. "lro on rxvlan off" setting them all together would cause (a regression to some case in the past) to not set e.g. rxvlan when lro apply fails.
Please adopt to use a get(&curr) and then foreach flag [ if curr.flag != want.flag) set } like we do it in the
other flags having a struct (ring, eee, ...).

mtomaschewski:
Applied via https://github.com/openSUSE/wicked/pull/718 adding the missed checks

PR #734, opened 2018-02-15
ibft: no IP setup on bnx2x storage-only interfaces (bsc#1072343) 

mtomaschewski:


mtomaschewski:
Depends on https://github.com/openSUSE/wicked/pull/733

Issue #735, opened 2018-02-20
wicked show-xml eth0' shows info about all interfaces (not filtering)


nirmoy:
'wicked show-xml eth0' it should display information about that specific network interface but instead it's showing info about all interfaces.

mtomaschewski:
Use e.g. ni_string_array_t as match filter (empty array == all), not just char *ifname to permit:

- "wicked show-xml eth0 eth1"
- "wicked show-xml" == "wicked show-xml all"

asmorodskyi:
this issue can be closed

PR #744, opened 2018-03-16
Fix show-xml filtering by interface name (issue #735,bsc#954758)

rtorrero:
This PR adds filtering functionality to the show-xml command, with the following behaviour:
  - `wicked show-xml <ifname1> <ifname2>`: show xml *only* for `ifname1` and `ifname2`
  - `wicked show-xml all`: show xml for all available network interfaces
  - `wicked show-xml`: same as above, show all interfaces

This should address issue #735

### DIFF
--- a/client/main.c
+++ b/client/main.c
@@ -523,7 +523,7 @@ __dump_fake_xml(const ni_dbus_variant_t *variant, unsigned int indent, const cha
 }
 
 static ni_bool_t
-__dump_object_xml(const char *object_path, const ni_dbus_variant_t *variant, 
+__dump_object_xml(const char *object_path, const ni_dbus_variant_t *variant,
 	ni_xs_scope_t *schema, xml_node_t *parent, const ni_string_array_t *filter)
 {
 	xml_node_t *object_node;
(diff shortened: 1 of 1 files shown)


rtorrero:
I'm not entirely happy with the double for loop, it looks messy. I'm open to suggestions if that can be improved.
If there is no better way to do it using any of the ni_dbus_variant or ni_dbus_dict functions, variable naming could use some love too :p.

mtomaschewski:
Please improve & fix as commented via pm mail :-)

rtorrero:
Thanks for your comments. Hope those commits address all the issues. Awaiting your feedback :).

client/main.c
```diff
@@ -522,50 +522,68 @@ __dump_fake_xml(const ni_dbus_variant_t *variant, unsigned int indent, const cha
...
 	if (!ni_dbus_variant_is_dict(variant)) {
 		ni_error("%s: dbus data is not a dict", __func__);
-		return NULL;
+		return FALSE;
 	}
 
-	object_node = xml_node_new("object", parent);
-	xml_node_add_attr(object_node, "path", object_path);
+	object_node = xml_node_new("object", NULL);
+	xml_node_add_attr(object_node, "path", object_path);
```
mtomaschewski: please remove the leading spaces. you may configure .gitconfig to show (and or fix) them, e.g. _core.whitespace = fix,indent-with-non-tab,trailing-space_

rtorrero:
I've added your additional suggestions (usage info and trailing whitespaces), but I did it before squashing the commits. Should I have done those afterwards?

mtomaschewski:
This looks very good now... There is still a bit odd usage inside of the show-xml, could you adjust it too please?

rtorrero:
I'm not sure how to improve that one, I'm looking at how other subcommands have their usage section and doing the ```< --raw | [ <ifname ...>|all ] >``` thing would mean that we have to remove the 'Supported options' part (as --raw is one of the options). Could you provide more details?

mtomaschewski:
Yes, your note should be sufficient to cover --raw case (until we refactor it to apply the filter to --raw too), but we should adjust at least:
```diff
- "wicked [options] show-xml <ifname|all>\n"
- "\nSupported options:\n"
+ "wicked show-xml [options] [ifname ... |all]\n"
+ "\n"
+ "Supported options:\n"
```

rtorrero:
I got you, should be done :). Thanks for the feedback!

mtomaschewski:
Could you also please also remove the obsolete variable? :
```
main.c: In function ‘do_show_xml’:
main.c:613:14: warning: unused variable ‘ifname’ [-Wunused-variable]
  const char *ifname = NULL;
              ^
```

mtomaschewski:
OK, we'll fix it in a separate commit.

PR #751, opened 2018-06-01
ethtool: streamline options available on all devices (bsc#1085786)

mtomaschewski:


rtorrero:
Noticed a few small typos in man/wicked-ethtool.8.in, nothing relevant, will correct later with a future PR

PR #755, opened 2018-06-08
documentation: various cleanups

rtorrero:
Since we have a new FAQ section on the wiki, it doesn't make much sense anymore to keep the old FAQ.txt in the documents.

I'd like some feedback, as I'm not sure it was the right call (would there be any case where the user could have the git repo checked out but no access to the wiki?)

I've also added a reference to the FAQ on the README and done some cleanups.

### DIFF
--- a/man/wicked-ethtool.8.in
+++ b/man/wicked-ethtool.8.in
@@ -88,10 +88,10 @@ Configure NIC's link related settings.
 .RS
 .TP
 .B autoneg <on|off>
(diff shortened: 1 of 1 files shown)


mtomaschewski:
I'd split this pull request. IMO we should evaluate a way to enable pull requests for the wiki, include the online-documentation (or it's output) in a source archive so we can ship it as wicked-doc subpackage and e.g. make manual pages available online. Something like that.

PR #757, opened 2018-07-31
man: add missing documentation for DHCLIENT6_CLIENT_ID

rtorrero:
This option was implemented a while ago but was not still documented in sysconfig/network/dhcp nor man pages.


mtomaschewski:
Please note, that this is a manual override option which also breaks an RFC4361 DHCPv4 client-id synchronization. Preferred way to set / change the duid using the "wicked duid" command.

PR #765, opened 2018-08-10
dhcp6: use timeval in ia and add list manipulation utils

mtomaschewski:


src/dhcp6/fsm.c
```diff
@@ -1825,12 +1825,11 @@ __ni_dhcp6_fsm_mark_ia_by_time(ni_dhcp6_device_t *dev,  unsigned int (*get_ia_ti
...
-		if ((aq = ia->time_acquired) == 0)
-			aq = dev->lease->acquired.tv_sec;
+		if (timercmp(&now, &ia->acquired, >)) {
+			struct timeval dif;
 
-		if (now.tv_sec > aq) {
-			diff = (now.tv_sec - aq);
-			if (diff + 1 >= rt) {
+			timersub(&now, &ia->acquired, &dif);
+			if (dif.tv_sec > rt) {
```
rtorrero: This if seems wrong and is causing renew to not happen, probably should be:
`if (dif.tv_sec + 1 >= rt)`

rtorrero:
LGTM

PR #774, opened 2018-10-01
firewall: do not assign default zone, but pass as is (bsc#1109147)

rtorrero:
This PR changes the firewall extension to not get the default zone from firewall-cmd for the --change-interface, but rather pass a empty zone string; firewalld should interpret this empty string as "use default zone".

### DIFF
--- a/extensions/firewall
+++ b/extensions/firewall
@@ -66,10 +66,7 @@ firewalld_up()
 	test "X$WICKED_INTERFACE_NAME" = "X" && return 1
 
 	local ZONE=`wicked_config_get_zone "$WICKED_ARGFILE"`
-	if test "X$ZONE" = "X" ; then
-		ZONE=`"$firewalld_cmd" --get-default-zone 2>/dev/null`
-		test "X$ZONE" = "X" && return 1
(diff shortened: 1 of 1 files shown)


extensions/firewall
```diff
@@ -66,10 +66,7 @@ firewalld_up()
 	test "X$WICKED_INTERFACE_NAME" = "X" && return 1
 
 	local ZONE=`wicked_config_get_zone "$WICKED_ARGFILE"`
-	if test "X$ZONE" = "X" ; then
-		ZONE=`"$firewalld_cmd" --get-default-zone 2>/dev/null`
-		test "X$ZONE" = "X" && return 1
-	fi
+
 	"$firewalld_cmd" --zone="$ZONE" --change-interface="$WICKED_INTERFACE_NAME" &>/dev/null
```
teclator: If firewalld permits a empty zone there as the default, for me looks good.

rtorrero: @hwoarang did a quick check on firewalld and it seems this behaviour is intentional, thanks for the reviews to both of you.

PR #775, opened 2018-10-02
nanny: fix memory leaks on fast create-delete calls (bsc#1095818)

mtomaschewski:


mtomaschewski:
there is a typo bug in commit 6d03c548f61c82d6b87891064544d0a003258ae5:
```
-       if (!mdev->object || dev_class)
+       if (!mdev->object || !dev_class)
```

PR #794, opened 2019-04-12
dhcp: request hostname/fqdn option in the tester

jvanz:
Back port the fix from SLES15 to SLES12 SP4


rtorrero:
Hi José,

As I told you on OBS, we want to push 0.6.54 to SLES12-SP* soon (not just cherry-picked parts of it). This should happen soon. Thanks!

PR #800, opened 2019-07-05
man: ifcfg-ovs-bridge(5): recommend STARTMODE=nfsroot

mwilck:
This may help avoiding user surprises.


mtomaschewski:
Yes, makes sense, this should avoid most of the questions we get about :-)

PR #801, opened 2019-07-08
rfkill: In switch statement check enum variable not constant

e4t:
This fixes broken Wifi support reported in:
https://bugzilla.suse.com/show_bug.cgi?id=1140117

Signed-off-by: Egbert Eich <eich@suse.com>


mtomaschewski:
Thanks, good catch!

PR #805, opened 2019-07-31
systemd: wait haveged to meet entropy requirements (bsc#1131369)

rtorrero:
Add a systemd (soft) dependency on havaged to wait for enough entropy without allowing this time to account as wicked start time.

### DIFF
--- a/etc/systemd/wickedd-auto4.service.in
+++ b/etc/systemd/wickedd-auto4.service.in
@@ -1,7 +1,7 @@
 [Unit]
 Description=wicked AutoIPv4 supplicant service
 Requisite=dbus.service
-After=local-fs.target dbus.service SuSEfirewall2_init.service
+After=local-fs.target dbus.service SuSEfirewall2_init.service haveged.service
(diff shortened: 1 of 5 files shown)


mtomaschewski:
This change is not needed -- it is given already:

* `systemd.special(7)` says:
  ```
       sysinit.target
           systemd automatically adds dependencies of the types Requires= and After= for this
           target unit to all services (except for those with DefaultDependencies=no).
  ```
* `haveged.service` is using:
  ```
  DefaultDependencies=no
  Before=sysinit.target shutdown.target systemd-journald.service
  ```
* `systemctl show wickedd.service | grep -E "^(Requires|After)="` proves that the deps to sysinit.target are in place:
  ```
  Requires=sysinit.target system.slice
  After=SuSEfirewall2_init.service isdn.service rdma.service systemd-udev-settle.service dbus.service system.slice systemd-journald.socket wickedd-dhcp6.service basic.target wickedd-auto4.service sysinit.target wickedd-dhcp4.service local-fs.target
  ```

PR #807, opened 2019-08-13
time: use boot time for timer instead of real time (bsc#1129986)

mtomaschewski:
Fixes to use boot time for timers/timeouts and real time only in lease
files to not stuck in timers due to adjustments of the system time.
Consistently log dhcp xid and enabled to log dhcp6 timings line.

### DIFF
--- a/src/wpa-supplicant.c
+++ b/src/wpa-supplicant.c
@@ -29,6 +29,7 @@
 #include "config.h"
 #endif
 
+#include <sys/time.h>
 #include <time.h>
 #include <net/if_arp.h>
 #include <netinet/if_ether.h>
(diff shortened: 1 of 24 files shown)


src/wpa-supplicant.c
```diff
@@ -622,7 +627,7 @@ ni_wpa_interface_retrieve_scan(ni_wpa_interface_t *wpa_dev, ni_wireless_scan_t *
 	if (ni_wpa_interface_expire_networks(wpa_dev, scan->interval + 1) == 0) {
 		/* Nothing pruned. If we didn't receive new scan results in the
 		 * mean time, there's nothing we need to do. */
-		if (scan->timestamp == wpa_dev->scan.timestamp)
+		if (!timercmp(&scan->timestamp, &wpa_dev->scan.timestamp, !=))
```
rtorrero: Is there any reason why you'd use a double negation here?

rtorrero:
After some testing wtih tcpdump to see if DHCPv4 was doing renewal in time, everything seems to be working correctly with this PR. LGTM!

src/wpa-supplicant.c
mtomaschewski: See "man timercmp", there are some implementation that don't work properly with '==':
```Some  systems  (but  not Linux/glibc), have a broken timercmp() implementation, in which CMP of >=, <=, and == do not work; portable applications can instead use
           !timercmp(..., <)
           !timercmp(..., >)
           !timercmp(..., !=)
```

rtorrero: You are totally right, marking as resolved

PR #808, opened 2019-08-14
dhcp6: do not default to a /64 address prefix-length (bsc#1132280)

mtomaschewski:
Add an address-length aka DHCLIENT6_ADDRESS_LENGTH ifcfg option, which
permits to specify explicit prefix-length to use for the DHCPv6 address,
e.g. 64 to use address as 2001:db8::1/64 or 80 for 2001:db8::1/80.
When 0 or unspecified (default), prefix-length of the smallest on-link
prefix (highest /length number) in the IPv6 router advertisement matching
the address is used or 128 (see also rfc5942).


rtorrero:
Did some quick test to confirm the behaviour:
  - when no RA are available, and the option is not specified, it defaults to /128
  - when option is specified, it uses the defined prefix in the option
  - when RA is available and option is not specified, it defaults to the lowest prefix available

LGTM, thanks!

PR #810, opened 2019-09-05
dhcp6: set the noprefixroute address option (bsc#1132280)

mtomaschewski:
Do not create an implicit prefix (route) by adding an address with prefix
other than 128, even when we've received an on-link RA prefix.

### DIFF
--- a/include/wicked/address.h
+++ b/include/wicked/address.h
@@ -129,11 +129,13 @@ extern ni_bool_t	ni_address_is_temporary(const ni_address_t *laddr);
 extern ni_bool_t	ni_address_is_permanent(const ni_address_t *laddr);
 extern ni_bool_t	ni_address_is_deprecated(const ni_address_t *laddr);
 extern ni_bool_t	ni_address_is_mngtmpaddr(const ni_address_t *laddr);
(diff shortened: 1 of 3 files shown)


rtorrero:
Related to: https://tools.ietf.org/html/rfc8415#section-18.2.10.1

Did a few tests to confirm it works when using `DHCLIENT6_MODE=managed` and `DHCLIENT6_ADDRESS_LENGTH=64` as well as `DHCLIENT6_ADDRESS_LENGTH=128` (to check if the kernel still likes `noprefixroute` in this case. Everything seems to be working as expected.

```
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether <removed> brd ff:ff:ff:ff:ff:ff
    inet6 <removed>/64 scope global noprefixroute dynamic 
       valid_lft 230sec preferred_lft 170sec
    inet6 <removed>/64 scope link 
       valid_lft forever preferred_lft forever

ruben@leap-test:~> ip -6 r s
fe80::/64 dev eth0 proto kernel metric 256 pref medium
ruben@leap-test:~> 
```

PR #819, opened 2019-12-09
dracut: expose cmdlines as config backend options

rtorrero:
This pull request adds initial support for parsing dracut's cmdline format so that wicked may be used to configure the network during the initramfs phase.


mtomaschewski:
regression test in the testsuite passed

PR #821, opened 2020-01-14
misc. bug fixes

maltek:
Fix a bunch of issues found through fuzzing and static analysis. Most should be self-explanatory.

The UAFs may sound bad at first, but can't be triggered by attackers according to my best judgement.

The weirder things are probably the fixes for undefined behavior: the standard is quite stupid there, but `memcpy` may not be called with a `NULL` parameter even if the length is `0`.

The most controversial change are probably those about alignment. I believe they are correct because the buffer API gives no alignment guarantees. However, in my fuzzer I initialized these buffers with memory that wasn't aligned to those types. It's quite possible that the buffers used by the actual network code are aligned in practice (thanks to malloc's alignment guarantees). If only for robustness reasons, I think these changes still make sense.

### DIFF
--- a/client/main.c
+++ b/client/main.c
@@ -1488,6 +1488,7 @@ do_check(int argc, char **argv)
 				if (opt_dbus_error_file) {
 					write_dbus_error(opt_dbus_error_file,
 							NI_DBUS_ERROR_UNRESOLVABLE_HOSTNAME,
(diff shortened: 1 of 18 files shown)


maltek:
Thanks, I've fixed the two commits.

PR #823, opened 2020-01-24
Packaging: BuildRequire pkgconfig(libsystemd) instead of systemd-devel

DimStar77:
On OBS/openSUSE this has the positive side-effect of using the -mini
flavors, meaning we can move up in fhe build queue and not be
blocked behind the 'full featured systemd package'. There is no
difference for the consumers.

### DIFF
--- a/wicked.spec.in
+++ b/wicked.spec.in
@@ -84,7 +84,7 @@ Requires(pre):       util-linux
 %endif
 
 %if %{with systemd}
-BuildRequires:  systemd-devel
+BuildRequires:  pkgconfig(libsystemd)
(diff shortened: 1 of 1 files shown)


rtorrero:
Thanks for the PR.

I just did a quick test and it doesn't seem to have any negative effect and I don't see a reason not to change it otherwise. WDYT @mtomaschewski ??

mtomaschewski:
> I just did a quick test and it doesn't seem to have any negative effect
No objections when it works this way also on SLE-12.

DimStar77:
> No objections when it works this way also on SLE-12.

pkgconfig() provides have been in place already in SLE-12-GA

PR #824, opened 2020-01-31
Fixes for startup order with systemd

### LINKED ISSUE #822: wickedd and openvswitch should not start in parallel
According to the official documentation [1], the OVS services need
to start long before the network (wicked) service, to be able to
interact with the service during boot. However, both OVS and wicked
are multi-service applications and the way that OVS and wicked systemd
service files are being set up, using a `Before=network.service` entry
in the OVS systemd unit file is not enough to guarantee that OVS
services are fully started before wicked services start.

Starting OVS and wicked in parallel can indeed cause problems
that are not easily reproducible, such as this:

```
Jan 22 23:35:12 dfa-16-3e-d2-b0-42 systemd[1]: Starting Open vSwitch...
Jan 22 23:35:13 dfa-16-3e-d2-b0-42 systemd[1]: Starting wicked network management service daemon...
Jan 22 23:35:13 dfa-16-3e-d2-b0-42 kernel: openvswitch: Open vSwitch switching datapath
Jan 22 23:35:13 dfa-16-3e-d2-b0-42 systemd[1]: Started wicked network management service daemon.
Jan 22 23:35:13 dfa-16-3e-d2-b0-42 systemd[1]: Starting wicked network nanny service...
Jan 22 23:35:13 dfa-16-3e-d2-b0-42 kernel: device br-fixed entered promiscuous mode
Jan 22 23:35:13 dfa-16-3e-d2-b0-42 systemd-udevd[981]: Could not generate persistent MAC address for br-fixed: No such file or directory
Jan 22 23:35:13 dfa-16-3e-d2-b0-42 ovs-ctl[797]: Starting ovs-vswitchd..done
Jan 22 23:35:13 dfa-16-3e-d2-b0-42 systemd[1]: Started Open vSwitch.
Jan 22 23:35:13 dfa-16-3e-d2-b0-42 systemd[1]: Started wicked network nanny service.
Jan 22 23:35:13 dfa-16-3e-d2-b0-42 systemd[1]: Starting wicked managed network interfaces...
Jan 22 23:35:13 dfa-16-3e-d2-b0-42 wickedd-nanny[947]: device br-public: failed to bind services and methods for waitDeviceReady()
Jan 22 23:35:13 dfa-16-3e-d2-b0-42 wickedd-nanny[947]: device br-fixed: failed to bind services and methods for waitDeviceReady()
```

This can easily be alleviated by customizing the OVS or wicked systemd service unit
configuration files, but I still think the official product should cover this somehow, either
by updating the documentation to say that the `Before=network.service` approach
is no longer valid, or by changing the way the wicked.service unit is synchronized
with the other service units to make that statement true again.

[1] https://en.opensuse.org/Portal:Wicked/OpenvSwitch

JanZerebecki:
Order wickedd After Open vSwitch on startup

Otherwise these are treading on each other while bringing up interfaces
which results in failures. Ordering Open vSwitch Before network.service
isn't suficient as wickedd and wickedd-nanny may start before that.

Fixes #822

---

Order After network-pre on startup

Add network-pre.target where ordering is already After
SuSEfirewall2_init.service.

SuSEfirewall2_init.service is also missing a Before on network-pre.
After that is done in SuSEfirewall2 it can be removed here.

Quote about network-pre from man systemd.special:
This passive target unit may be pulled in by services that want to run
before any network is set up, for example for the purpose of setting up
a firewall. All network management software orders itself after this
target, but does not pull it in.

### DIFF
--- a/etc/systemd/wickedd-auto4.service.in
+++ b/etc/systemd/wickedd-auto4.service.in
@@ -1,7 +1,7 @@
 [Unit]
 Description=wicked AutoIPv4 supplicant service
 Requisite=dbus.service
-After=local-fs.target dbus.service SuSEfirewall2_init.service
+After=local-fs.target dbus.service network-pre.target SuSEfirewall2_init.service
 Before=wickedd.service wicked.service network.target
 PartOf=wickedd.service
 
--- a/etc/systemd/wickedd-dhcp4.service.in
+++ b/etc/systemd/wickedd-dhcp4.service.in
@@ -1,7 +1,7 @@
 [Unit]
 Description=wicked DHCPv4 supplicant service
 Requisite=dbus.service
-After=local-fs.target dbus.service SuSEfirewall2_init.service
+After=local-fs.target dbus.service network-pre.target SuSEfirewall2_init.service
 Before=wickedd.service wicked.service network.target
 PartOf=wickedd.service
 
--- a/etc/systemd/wickedd-dhcp6.service.in
+++ b/etc/systemd/wickedd-dhcp6.service.in
@@ -1,7 +1,7 @@
 [Unit]
 Description=wicked DHCPv6 supplicant service
 Requisite=dbus.service
-After=local-fs.target dbus.service SuSEfirewall2_init.service
+After=local-fs.target dbus.service network-pre.target SuSEfirewall2_init.service
 Before=wickedd.service wicked.service network.target
 PartOf=wickedd.service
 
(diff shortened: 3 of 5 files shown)


rtorrero:
Thanks for the PR, LGTM.

As side note: even though 17e60f2cce44b9ad267b3aed3b17b2808b133e23 adds something we should have had according to https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/, in this case it is not enough to address #822, as network-pre.target is too soon for openvswitch.

I'll test on monday/tuesday to see if there is are any regressions and merge otherwise

rtorrero:
Passes openQA tests, also did some manual tests with SuSEfirewall2 on SLE12* and it works as expected.

PR #825, opened 2020-02-05
schema: change rto_min key to avoid not applying it (bsc#1160939)

rtorrero:
Two issues prevent the `rto_min` value from being applied on a route when following the `ip route` syntax, e.g.:

`ip route add 172.22.151.0/24 via 172.22.99.3 dev enp1s0 rto_min 20ms`

which in our `ifroute-xxx` syntax means:
```
localhost:~ # cat /etc/sysconfig/network/ifroute-enp8s0 
default	172.22.100.1	-	-	rto_min 20ms
```

1) we expect `rto_min` value to be an int, so the `ms` (as in the default `ip` syntax) causes that it doesn't get read correctly.

2) when it is just the numeric value, we parse it and store it in our schema as `rto-min`. In our dbus layer, before sending it to the kernel, we instead attempt to load `rto_min`, resulting in the setting never being applied.

This PR fixes the latter.

### DIFF
--- a/client/compat.c
+++ b/client/compat.c
@@ -1801,7 +1801,7 @@ __ni_compat_generate_static_route_metrics(xml_node_t *mnode, const ni_route_t *r
 		xml_node_new_element("features", mnode, ni_sprint_uint(rp->features));
 	}
 	if (rp->rto_min > 0) {
-		xml_node_new_element("rto-min", mnode, ni_sprint_uint(rp->rto_min));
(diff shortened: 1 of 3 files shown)


mtomaschewski:
The `s`, `sec`, `ms`, `msec` or `msecs` suffix is a new addition to ip-route as mentioned in it's man page. Could you also please parse the suffix from `ifroute(5)` files for `rtt`, `rttvar`, `rto_min` (and other values) using the time format?

PR #827, opened 2020-02-12
Introduce .editorconfig file

cfconrad:
Convenient way to have project based editor config, see http://EditorConfig.org

I walked through the project and tried to identify the most common used
ways for different file types.

Following four types:

  * xml and .conf files
    => 2 spaces
  * manpages
    => 3 spaces
  * doc/html
    => 1 space
  * all others
    => tab with 8

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,28 @@
+# EditorConfig: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+insert_final_newline = true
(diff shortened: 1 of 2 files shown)


cfconrad:
@mtomaschewski WDYT?

mtomaschewski:
> Convenient way to have project based editor config, see http://EditorConfig.org
>     * doc/html
>       => 1 space

The html pages are generated from schema ... but it does not hurt.

PR #836, opened 2020-06-17
Don't block network.target with wicked (boo#1172684)

lnussel:
network.target is meant to activate networking daemons but not block
waiting for interfaces to come up. That's what network-online.target
is for. So adjust service files to hook wickedd.service into
network.target and wicked.service into network-online.target.

### DIFF
--- a/etc/systemd/wicked.service.in
+++ b/etc/systemd/wicked.service.in
@@ -2,7 +2,7 @@
 Description=wicked managed network interfaces
 Wants=network.target wickedd.service wickedd-nanny.service
 After=network-pre.target wickedd.service wickedd-nanny.service
-Before=network-online.target network.target
+Before=network-online.target
 
 [Service]
 Type=oneshot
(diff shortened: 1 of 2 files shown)


lnussel:
works on rpi2 with those service files, didn't try upgrade case though

mtomaschewski:
I don't think we can do it this way at all without to rewrite yast2 & co + whole upgrade path, but we'll see. It needs a very careful testing ...

lnussel:
as long as yast doesn't maintain the network symlink itself and just uses systemctl to enable the units it shouldn't even bother. I can open a feature request if that helps

mtomaschewski:
> as long as yast doesn't maintain the network symlink itself and just uses systemctl to enable the units it shouldn't even bother. I can open a feature request if that helps

A feature requests are maybe needed, but I currently don't have an overview what all would be needed to adopt by applying your service changes. We maintain **one** wicked for all versions,
so it has to work for all maintained versions (and match docs, ...).
But as I wrote -- it needs a closer review of all involved parts and this can't be done out of order, but needs to be scheduled properly.

lnussel:
Ah so that is where the claim that SLE impairs advances in TW comes from :-( I'm sure there must be a way to be able to give this a shot in TW first.

lnussel:
jsc#PM-1982

mtomaschewski:
> jsc#PM-1982

Thanks!

PR #837, opened 2020-07-01
config: fix include file name buffer handling

mtomaschewski:


mtomaschewski:
When the /etc/wicked/local.xml exists, wicked client (e.g. `wicked test dhcp4 eth0`) were including:
```
   wicked: Reading config file /etc/wicked/client.xml
   wicked: Reading config file /etc/wicked/common.xml
```
but the 2nd level local.xml include file weren't due to this buffer usage bug:
```
   wicked: Reading config file /etc/wicked/local.xml
``

PR #839, opened 2020-07-29
teamd: fix schema to use correct hwaddr_policy property (boo#1171234)

rtorrero:
Schema defines that `hwaddr-policy` is where we store `TEAM_AB_HWADDR_POLICY` but client/compat.c is using `hwaddr_policy` instead, resulting in:

`"wickedd-nanny[710]: ni_dbus_serialize_xml_dict: ignoring unknown dict element "hwaddr_policy"`

This PR fixes the schema to use hwaddr_policy as in src/dbus-objects/team.c, client/compat.c and teamd.conf(5).

### DIFF
--- a/schema/team.xml
+++ b/schema/team.xml
@@ -15,7 +15,7 @@
   </define>
 
   <define name="activebackup-runner" class="dict">
-    <hwaddr-policy     type="builtin-team-ab-hwaddr-policy" />
+    <hwaddr_policy     type="builtin-team-ab-hwaddr-policy" />
   </define>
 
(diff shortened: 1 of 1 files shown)


rtorrero:
Adjusted. Initially my priority was to leave the schema untouched but I realize that consistency with the rest of the code is probably more relevant here.

mtomaschewski:
> consistency with the rest of the code is probably more relevant here.
Not only with the rest of the code, but also with `man 5 teamd.conf` -- we are able and are using exactly same property names, so it makes sense to keep doing it.

PR #843, opened 2020-09-21
dhcp4: add DHCLIENT_CREATE_CID to ifcfg (jsc#SLE-15770)

mtomaschewski:
Overrides the DHCPv4 client-identifier type to use specified in the
wicked-config(5) `create-cid` option, the interface type specific
client-id type and compile time defaults.
Note: DHCP over Infiniband (IPoIB) mandates an rfc4361 client-id.

### DIFF
--- a/client/compat.c
+++ b/client/compat.c
@@ -2236,6 +2236,9 @@ __ni_compat_generate_dhcp4_addrconf(xml_node_t *ifnode, const ni_compat_netdev_t
 
 	if (compat->dhcp4.client_id)
 		xml_node_dict_set(dhcp, "client-id", compat->dhcp4.client_id);
+	else
+	if (compat->dhcp4.create_cid)
+		xml_node_dict_set(dhcp, "create-cid", ni_config_dhcp4_cid_type_format(compat->dhcp4.create_cid));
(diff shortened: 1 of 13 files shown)


rtorrero:
Shouldn't we also add a description for DHCLIENT_CREATE_CID in `man ifcfg-dhcp` ?

mtomaschewski:
> Shouldn't we also add a description for DHCLIENT_CREATE_CID in `man ifcfg-dhcp` ?

Yes, done. And fixed the issues (we've never checked result of the parsing function before...).

rtorrero:
Yes, I noticed that DHCLIENT_CREATE_CID was not being parsed correctly. Going to re-test now.

> > Shouldn't we also add a description for DHCLIENT_CREATE_CID in `man ifcfg-dhcp` ?
> 
> Yes, done. And fixed the issues (we've never checked result of the parsing function before...).

rtorrero:
OK, iIve retested with tcpdump & wireshark and it works perfectly now. Also the man page section is there now. LGTM, thanks!

rtorrero (approved):
LGTM

PR #844, opened 2020-09-24
Avoid incomplete ifdown/timeout on route deletion error (bsc#1174099)

rtorrero:
This MR bsc#1174099 by allowing an ifdown to continue when the route marked for deletion was already deleted.


mtomaschewski (approved):
OK, thanks!

PR #845, opened 2020-09-29
ifreload: check running slaves on master reload (bsc#1168155,bsc#1172082)

mtomaschewski:
Check if config of currently enslaved interfaces changed also in `wicked ifreload brX` to reload (shutdown) slaves without config and removed from the slave/port list.


rtorrero (approved):
LGTM & passed all tests, thanks!

PR #850, opened 2020-11-24
ifup: fix for set up with unenslave from unconfigured master (boo#954329)

mtomaschewski:


src/fsm.c
```diff
@@ -4104,7 +4108,7 @@ __ni_ifworker_refresh_modems(ni_fsm_t *fsm)
 	}
 
 	for (object = list_object->children; object; object = object->next) {
-		ni_fsm_recv_new_modem(fsm, object, FALSE);
+		ni_fsm_recv_new_modem(fsm, object, TRUE);
```
cfconrad, line 4111: From the code "pattern" I'm missing here https://github.com/openSUSE/wicked/blob/468a60bcf107b7375702490a4a1171860429d852/src/fsm.c#L4125 a 
```
if (modem) ni_modem_reset(modem);
```
WDYT?

src/netdev.c
```diff
@@ -94,52 +94,71 @@ ni_netdev_slaveinfo_destroy(ni_slaveinfo_t *slave)
...
+	dev->link.metric = 0;
+	dev->link.txqlen = 0;
+	dev->link.oper_state = 0;
+
 	ni_netdev_ref_destroy(&dev->link.lowerdev);
 	ni_netdev_ref_destroy(&dev->link.masterdev);
 	ni_netdev_slaveinfo_destroy(&dev->link.slave);
+
+	ni_string_free(&dev->link.qdisc);
+	ni_string_free(&dev->link.kind);
```
cfconrad, line 124: Maybe it makes sense to move these in a `ni_linkinfo_reset()`?

cfconrad: what about `dev->link.saved_mtu`?

src/netdev.c
```diff
@@ -94,52 +94,71 @@ ni_netdev_slaveinfo_destroy(ni_slaveinfo_t *slave)
...
-	/* Clear out addresses, routes, ... */
 	ni_netdev_clear_addresses(dev);
 	ni_netdev_clear_routes(dev);
-	ni_netdev_set_ethernet(dev, NULL);
-	ni_netdev_set_infiniband(dev, NULL);
-	ni_netdev_set_bonding(dev, NULL);
+
+	ni_netdev_set_ipv4(dev, NULL);
+	ni_netdev_set_ipv6(dev, NULL);
+	ni_netdev_set_auto6(dev, NULL);
```
cfconrad, line 134: Side note:
https://github.com/openSUSE/wicked/blob/468a60bcf107b7375702490a4a1171860429d852/src/auto6.c#L171
never set anything, it always do a free...

src/netdev.c
```diff
@@ -94,52 +94,71 @@ ni_netdev_slaveinfo_destroy(ni_slaveinfo_t *slave)
...
+ni_netdev_reset(ni_netdev_t *dev)
 {
-	/* Clear out linkinfo */
-	ni_string_free(&dev->link.qdisc);
-	ni_string_free(&dev->link.kind);
+	if (!dev)
+		return;
+
+	/* Clear and Reset all (kernel) property data,
+	 * except of list, ref counter & marker flags */
```
cfconrad, line 104: Just curious, what about:
```
dev->seq
dev->modifed
dev->deleted
dev->created

src/fsm.c
mtomaschewski: Did you noticed the `#ifdef MODEM` around these functions that disables incomplete modem functions/support completely?

src/netdev.c
mtomaschewski: dev->link.saved_mtu is an internal thing used in wickedd only, not an exposed property.

I were thinking about `ni_linkinfo_reset()`, but it is not needed anywhere and it is IMO quite useless to reset &dev->link (= part of dev) without to reset `dev->*`.

src/netdev.c
mtomaschewski: Well, true. There is no config to assign (should be `ni_auto6_conf_t *conf`) like the sysctl's in ipv4/ipv6 -- it's currently
state only always filled by kernel events.
I didn't changed any of these functions, just their call order to match order in the struct to ensure we don't miss to reset
any struct member (and cause a leak). When, then it is a separate thing to cleanup next time we touch auto6.

src/fsm.c
cfconrad: ok, where do I find the the `define`'s?

src/netdev.c
mtomaschewski: This is same as with saved_mtu -- it's a state flag we use when we process events in **wickedd**
and not a property (e.g. when we send NEWLINK creating a new  device, we set dev->created
until a NEWLINK event from kernel arrives that "updates" it -> dev->created causes to send a
NI_EVENT_DEVICE_CREATE dbus event then in contrast to NI_EVENT_DEVICE_CHAGED
or other events --> all arriving as NEWLINK from kernel).

The reset is done in the fsm (nanny/client side): we need to reset (ni_netdev_t)dbus-object->handle
before we apply properties to it. When wickedd sends us a message with e.g. master "br0" reference
property, we set the dev->link.masterdev.name/index. Then it gets unenslaved in kernel, there is a
NEWLINK which resets  dev->link.masterdev.name/index. In the dbus layer, the dbus-propery->get()
returns no_property for "master" property in the client side msg does not contain any -- this caused,
that the dev->link.masterdev.name/index were not reset from "br0" to NULL.

src/fsm.c
mtomaschewski: e.g. in https://github.com/openSUSE/wicked/blob/468a60bcf107b7375702490a4a1171860429d852/src/fsm.c#L4092
There is no `AC_ARG_ENABLE([modem], ...)` + `AC_DEFINE([MODEM], ...` in  configure.ac that would permit to enable it -> dead code.

cfconrad: thx for explanation!

src/netdev.c
mtomaschewski: dev with link.ifindex == 0 is not a valid device (in the kernel or in wickedd).

src/netdev.c
cfconrad: ic, thx!

cfconrad (approved):
LGTM

PR #852, opened 2020-11-26
dbus: add initial variant container type handling

cfconrad:
A dbus variant container with inner (typed) variant is required for migration to wpa_supplicant WPA1 dbus API.


mtomaschewski (approved):
OK, thanks! Looks good and works for me.

PR #853, opened 2020-11-27
Let wicked-service also provide service(network)

Vogtinator:
The old provides is kept for compatibility for now.

On a related note, `Requires(pre): %name` looks wrong here. It means that `wicked` can be removed while `wicked-service` is still happily installed.


rtorrero:
@Vogtinator Thanks for the PR. And also for the comment about `Requires(pre): %name`, we will take a look at that on a separate internal issue.

rtorrero (approved):

PR #854, opened 2020-11-30
prepare usrmerge (boo#1029961)

lnussel:
In order to prepare for Tumbleweed to finish the [UsrMerge](https://en.opensuse.org/openSUSE:Usr_merge), the compat symlinks in /(s)bin have to get out of the way. The rpm macro %usrmerged was introduced to allow to prepare packages and to be able to do the switch globally once as soon as all packages are prepared.

### DIFF
--- a/wicked.spec.in
+++ b/wicked.spec.in
@@ -201,14 +201,18 @@ make %{?_smp_mflags}
 
 %install
 make install DESTDIR=${RPM_BUILD_ROOT}
+%if !0%{?usrmerged}
 # install /sbin/{ifup,ifown,ifstatus,ifprobe} links
-%if "%_sbindir" != "/sbin"
 %__mkdir_p -m 0755 ${RPM_BUILD_ROOT}/sbin
 %__ln_s %_sbindir/ifup	${RPM_BUILD_ROOT}/sbin/ifup
 %endif
-%__ln_s %_sbindir/ifup	${RPM_BUILD_ROOT}/sbin/ifdown
-%__ln_s %_sbindir/ifup	${RPM_BUILD_ROOT}/sbin/ifstatus
-%__ln_s %_sbindir/ifup  ${RPM_BUILD_ROOT}/sbin/ifprobe
+for i in ifdown ifstatus ifprobe; do
+%if !0%{?usrmerged}
+%__ln_s ifup ${RPM_BUILD_ROOT}/sbin/$i
+%else
+%__ln_s ifup ${RPM_BUILD_ROOT}%{_sbindir}/$i
+%endif
+done
 # remove libwicked.a and la
 %__rm -f ${RPM_BUILD_ROOT}%_libdir/libwicked*.*a
 # create reboot-persistent (leases) store directory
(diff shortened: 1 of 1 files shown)


mtomaschewski (changes requested):
A  pull request which not even provides a short describtion of the intention? Is  the pull not worth to document or not worth to consider at all?

Wicked installs all binaries to `/usr/sbin` **since SLE-12-GA** (and before) -- inclusive of `ifup` wrapper script.
The `/sbin/if{up,down,status,probe}` symlinks and the wrapper script are provided for sysconfig aka **SLE-11 compatibility**, where the scripts were in `/sbin`. It is rather an coincidence, that there were also the *usrmerge* initiative moving things not required at boot to `/usr`.

There is code that is using  the `/sbin/ifup` compatibility wrapper, e.g. `execv("/sbin/ifup", ...),
refering to it, ... which would start to fail without it.
Thus  as long as  we provide such an compatibility, they have to stay in  `/sbin`. When we don't provide the compatibility,  IMO the complete wrapper script and the links can be removed.

mtomaschewski:
I need a PM approval and thus a feature request in order to remove the SLE-11 `ifup` compatibility -- this is not a bug to provide it and thus the bug is simply invalid.

mtomaschewski:
IMO a right way to implement a removal of the SLE-11 `ifup` compatibility would be to use conditionals _like_ `%bcond_without sle11_ifup_compat` and `%bcond_with sle11_ifup_compat`, conditionally move them to a `wicked-sle11-ifup-compat`, ...

lnussel:
the code doesn't remove anything as long as %usrmerge isn't set and that is not the case anywhere yet. So a total noop for released distros

lnussel:
I've tweaked the pr to make it even more clear what's  going on.

lnussel:
Any comment on the updated pr?

mtomaschewski:
Yes: the code changes neither changed our priorities about this issue nor anything else that I've mentioned before.
I'll not accept an external variable causing to build RPMs breaking sle11 compatibility once you decided to define this flag somewhere.

lnussel:
You lost me here. From what I can see nothing breaks for sle11. %usrmerged will never be set there. It's to be defined in Factory once all packages are prepared. See also
https://en.opensuse.org/openSUSE:Usr_merge#Preparation

lnussel:
Wait, you mean your concern is that `/sbin/ifup` won't exist anymore after `%usrmerged ` is set? That's not the case. After usr***merge*** both `/sbin` and `/usr/sbin` refer to the same location. So even though `ifup` will be installed as `/usr/sbin/ifup`, calling `/sbin/ifup` will still work as `/sbin` is a symlink to `/usr/sbin` (just like `bin`, `lib` and `lib64`).

mtomaschewski (approved):
> Wait, you mean your concern is that `/sbin/ifup` won't exist anymore after `%usrmerged ` is set? [...] calling `/sbin/ifup` will still work as `/sbin` is a symlink to `/usr/sbin` (just like `bin`, `lib` and `lib64`).

OK, `/sbin` will be a symlink to `/usr/sbin` and  there is a clear description of the intention of the pull request with reference to the [UsrMerge](https://en.opensuse.org/openSUSE:Usr_merge) page.

lnussel:
What's the ETA for landing this in Factory?

PR #855, opened 2021-01-21
 dbus: Display element name on unknown enum value error

cfconrad:
If a enum value contain unknown things, the error message should output
also the element name which it belongs to.
New output format looks like:

```
Dec 04 13:57:54 linux-22ob wickedd-nanny[3372]: /run/wicked/nanny/policy__wlan1.xml:21: element <auth-proto>: unknown enum value "wpa1,wpa2"
```


mtomaschewski (approved):
LGTM

PR #856, opened 2021-02-04
client: parse sysctl files in the correct order (bsc#1181186)

Dzejrou:
The files in sysconfig directories are often prefixed with numbers to indicate an order. Presently, wicked ignores that order and so some settings that are expected to be applied last may be applied early and then overriden (e.g. by some default).

This PR adds sorting to the list of sysconfig files, fixes bsc#1181186.


mtomaschewski:
Thanks! I've changed sysconfig to sysctl files in subject. I'll put it into testing this week.

cfconrad (approved):
LGTM, thx!

PR #857, opened 2021-02-15
ifconfig: differentiate if to re-trigger dad on address updates (bsc#1177215)

mtomaschewski:
Re-trigger duplicate address detection when we may have moved to a new link, but skip it on
normal lifetime update / renew when it has been already performed (RFC8415, 18.2.10.1) to
not cause 1s packet loss during the detection.


rtorrero:
Testsuite passes all tests. Manually tested the change on leap15.2, seems to work as expected. Thanks!

rtorrero (approved):

PR #858, opened 2021-02-23
dbus: Add generic object_path and object_path_array property macros

cfconrad:
This commit add the two macros:

  * `NI_DBUS_GENERIC_OBJECT_PATH_PROPERTY`
  * `NI_DBUS_GENERIC_OBJECT_PATH_ARRAY_PROPERTY`

and the required impelementations to fill `.set`, `.get` and `.parse` in
the ni_dbus_property_t structure.


mtomaschewski (approved):
LGTM.

BTW:  We should look and get rid of the .parse function pointer -- it is IMO never used / obsoleted by notations in schema.

PR #861, opened 2021-03-09
[dbus] Add generic property for `byte_array`

cfconrad:
This add a generic property handling for the DBus type `ab` (ByteArray), the macro is `NI_DBUS_GENERIC_BYTE_ARRAY_PROPERTY`. It will be mapped to the generic `ni_byte_array_t` type.

The type `ni_byte_array_t` is added with all needed utility functions like new, init, destroy, free, size and put[sb].


mtomaschewski (approved):
LGTM.

PR #862, opened 2021-03-10
client: expose ethtool --get-permanent-address option

mtomaschewski:


rtorrero:
Testsuite passed & manual tests seem to show it works as expected. Thanks!

rtorrero (approved):
Testsuite passed & manual tests seem to show it works as expected. Thanks!

PR #864, opened 2021-03-15
Remove systemd-udev-settle

madanyang:
According to udevadm systemd-udev-settle.service is deprecated. Hence removed from Wants and and After

### DIFF
--- a/etc/systemd/wickedd.service.in
+++ b/etc/systemd/wickedd.service.in
@@ -1,8 +1,8 @@
 [Unit]
 Description=wicked network management service daemon
 Requisite=dbus.service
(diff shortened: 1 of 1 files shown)


mtomaschewski:
Yes, but it has to wait until sle15-sp3 were released and we've retested the behavior without it.

mrjimenez:
> Yes, but it has to wait until sle15-sp3 were released and we've retested the behavior without it.

That's been a while ago, how about dropping systemd-udev-settle now?

Regards,
Marcelo.

mtomaschewski (approved):

PR #865, opened 2021-03-31
dhcp4: discover on reboot timeout after start-delay (bsc#1181812)

mtomaschewski:
When enabled, randomize start delay in range 1..config, max 10sec.


rtorrero (approved):
Testsuite passed & manual test shows functionality to be working correctly, thanks!

PR #866, opened 2021-03-31
dhcp6: request nis options on sle15 by default (bsc#1181812)

mtomaschewski:


mtomaschewski:
I've just fixed a change in commit c2bc6a9b, that was wrong/too much:
```
diff --git a/src/names.c b/src/names.c
index e566a243..83164552 100644
--- a/src/names.c
+++ b/src/names.c
@@ -305,7 +305,6 @@ ni_addrconf_update_flags_parse_names(unsigned int *flags, const ni_string_array_
        if (!flags || !names)
                return FALSE;
 
-       mask = *flags;
        for (i = 0; i < names->count; ++i) {
                if (!(name = names->data[i]))
                        continue;
```
It caused to always merge "default" set together with the requested update flags and e.g. `DHCLIENT[6]_UPDATE="dns"` contained also other flags from default set instead to contain only dns flag set.

rtorrero (approved):
testsuite passed, manual test on SLE15-SP2 -> DHCPv6 Solicit now shows additional requested Options 27 (NIS) and 29 (NIS Domain Name).

Thanks for the cleanup too!

PR #867, opened 2021-04-09
utils: format/parse bitmap to array and string alternatives

cfconrad:
Add bitmap utility functions:
 * `ni_parse_bitmap_array()`
 * `ni_parse_bitmap_string()`
 * `ni_format_bitmap_array()`
 * `ni_format_bitmap_string()`

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -1582,18 +1582,11 @@ ni_string_split(ni_string_array_t *nsa, const char *str, const char *sep,
 const char *
 ni_string_join(char **str, const ni_string_array_t *nsa, const char *sep)
 {
-	ni_stringbuf_t buf;
-	unsigned int i;
+	ni_stringbuf_t buf = NI_STRINGBUF_INIT_DYNAMIC;
 
-	if (nsa == NULL || sep == NULL || str == NULL)
+	if (!sep || !ni_stringbuf_join(&buf, nsa, sep))
 		return NULL;
 
-	ni_stringbuf_init(&buf);
-	for (i=0; i < nsa->count; ++i) {
-		if (i)
-			ni_stringbuf_puts(&buf, sep);
-		ni_stringbuf_puts(&buf, nsa->data[i]);
-	}
 	ni_string_dup(str, buf.string);
 	ni_stringbuf_destroy(&buf);
 
@@ -2253,6 +2246,24 @@ ni_bool_t
 ni_stringbuf_empty(const ni_stringbuf_t *sb)
 {
 	return sb->len == 0; /* bool */
+}
+
+const char *
+ni_stringbuf_join(ni_stringbuf_t *buf, const ni_string_array_t *nsa, const char *sep)
+{
+	unsigned int i;
+	size_t len;
+
+	if (!buf || !nsa)
+		return NULL;
+
+	len = buf->len;
+	for (i = 0; i < nsa->count; ++i) {
+		if (sep && buf->len)
+			ni_stringbuf_puts(buf, sep);
+		ni_stringbuf_puts(buf, nsa->data[i]);
+	}
+	return buf->string ? buf->string + len : NULL;
 }
 
 inline static size_t
--- a/src/util.c
+++ b/src/util.c
@@ -2156,11 +2156,36 @@ ni_parse_bitmap_string(unsigned int *mask, const ni_intmap_t *map, const char *i
 	return err;
 }
 
-const char *
-ni_format_bitmap(ni_stringbuf_t *buf, const ni_intmap_t *map,
-		unsigned int flags, const char *sep)
+unsigned int
+ni_format_bitmap_array(ni_string_array_t *array, const ni_intmap_t *map,
+		unsigned int mask, unsigned int *done)
 {
-	unsigned int i, flag;
+	unsigned int flag;
+
+	if (!array || !map)
+		return -1U;
+
+	for (; map->name; ++map) {
+		flag = NI_BIT(map->value);
+		if (mask & flag) {
+			if (ni_string_array_append(array, map->name) < 0)
+				continue;
+
+			mask &= ~flag;
+			if (done)
+				*done |= flag;
+		}
+	}
+
+	return mask;
+}
+
+const char *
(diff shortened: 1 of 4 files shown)


cfconrad:
As discussed, I changed the return values, `-1` on failure and `0` on success or `NULL` on failure and `char*` on success.

Also all "output" data get initialized before written.

Added a small test, to check all functionalities.

mtomaschewski (changes requested):
No, it seems that there was an misundertanding. Returning `unsigned int` count values as `int` is IMO always a bug. They **should** return `unsigned int` and `-1U`.

mtomaschewski:
What I meant before, I'd do it like in https://github.com/mtomaschewski/wicked/tree/bitmap-utils and return the
error count (success count is already in array.count increment) same as the parsing utils.

The `ni_stringbuf_join` returns pointer to the joined / constructed string -- when it was empty before, there is
no difference, but when not, it appends to the existing one, so something like this should also work:
```
   ni_stringbuf_puts(buf, "Flags: "); ni_stringbuf_join(buf, &flags, "|")
```
so it works like ```ni_stringbuf_puts(buf, "Flags: "); ni_stringbuf_printf(buf, "0x%x", 42);``` does.
Further,  it does not enforce to specify a separator, but omits it if NULL.

Everything is _untested_ again, except that it compiles ;-)

src/util.c
```diff
@@ -2255,6 +2320,24 @@ ni_stringbuf_empty(const ni_stringbuf_t *sb)
...
+	if (!buf || !nsa)
+		return NULL;
+
+	len = buf->len;
+	for (i = 0; i < nsa->count; ++i) {
+		if (sep && buf->len)
+			ni_stringbuf_puts(buf, sep);
+		ni_stringbuf_puts(buf, nsa->data[i]);
+	}
+	return buf->string ? buf->string + len : NULL;
```
cfconrad, line 2338: This is different from yours https://github.com/mtomaschewski/wicked/blob/bitmap-utils/src/util.c#L2243 but I found this better.

src/util.c
```diff
@@ -2119,28 +2112,100 @@ ni_parse_hex(const char *string, unsigned char *data, unsigned int datasize)
...
+			if (ni_string_array_append(array, map->name) < 0)
+				continue;
+
+			mask &= ~flag;
+			if (done)
+				*done |= flag;
+		}
+	}
+
+	return mask;
```
cfconrad, line 2180: This also differs from your version  https://github.com/mtomaschewski/wicked/blob/bitmap-utils/src/util.c#L2084

But I think in your version it was broken and actually returned if one flag of `map` wasn't set at all.
The other possibility is returning 0 here as well and just let the callee get the wrong bits with
```
unknown_bit_s = done ^ mask;

File: src/util.c [hunk 3]
```diff
@@ -2119,28 +2112,100 @@ ni_parse_hex(const char *string, unsigned char *data, unsigned int datasize)
...
+	ptr = ni_stringbuf_join(buf, &array, sep);
+	ni_string_array_destroy(&array);
+	return ptr;
+}
+
+const char *
+ni_format_bitmap(ni_stringbuf_t *buf, const ni_intmap_t *map,
+		unsigned int mask, const char *sep)
+{
+	return ni_format_bitmap_string(buf, map, mask, NULL, sep);
```
cfconrad, line 2208: I checked the calles of the following functions:
```
  ni_addrconf_flags_format(ni_stringbuf_t *buf, unsigned int flags, const char *sep)
  ni_addrconf_update_flags_format(ni_stringbuf_t *buf, unsigned int flags, const char *sep)
  ni_dhcp6_mode_format(ni_stringbuf_t *buff, unsigned int mask, const char *sep)
  ni_ethtool_wol_flags_format(ni_stringbuf_t *buf, unsigned int mask, const char *sep)
```
All of these looks OK to me!

src/util.c
```diff
@@ -2119,28 +2112,100 @@ ni_parse_hex(const char *string, unsigned char *data, unsigned int datasize)
...
-			ni_stringbuf_puts(buf, map->name);
-		}
-	}
-	return buf->string;
+	if (ni_format_bitmap_array(&array, map, mask, done) == -1U)
+		return NULL;
+
+	ptr = ni_stringbuf_join(buf, &array, sep);
+	ni_string_array_destroy(&array);
+	return ptr;
```
cfconrad, line 2201: Here I'm not sure about one case see this example:
```c
ni_stringbuf_destroy(sb);
ni_format_bitmap(sb, map, mask_in, "|"); 
// return the string "GET" => OK
ni_format_bitmap(sb, map, mask_in, "|"); 
// return the string "|GET" cause the `sb` stringbuffer isn't empty.
```
Not sure if we should include the first separator in the resulting string, when the given sb isn't empty.

cfconrad (commented):
Next try -- done from my side

src/util.c
mtomaschewski: Yes, the changes all make sense.

src/util.c
mtomaschewski: Yes, it was broken an counting valid but unset bits as errors -- yours is fine and returns the unknown bits as "error".
The format_string is less strict and returns NULL only on invalid args (-1U) and not all errors (unknown bits), but this is probably OK to format what we can/understand and discard any unknown/unnamed bits.

src/util.c
mtomaschewski: OK, thanks!

mtomaschewski:
There are two things left:
- Use tab (\t) to indent ni_parse_bitmap_array in include/wicked/util.h
- Don't use wrong/old files as example, but correct ones for copyright & co in bitmap-test.c:
```
[...]
 *      Copyright (C) 2021 SUSE LLC
[...]
 *      This program is free software; [...] see <http://www.gnu.org/licenses/>.
[...]
 *      Authors:
 *              Clemens Famulla-Conrad <cfamullaconrad@suse.com>
[...]
```

src/util.c
cfconrad: @mtomaschewski something to add here?

mtomaschewski (approved):

PR #870, opened 2021-04-15
xml-schema: add hex-string byte-array notation

cfconrad:
On XML, the data is in the format of `^([0-9a-f]{2})*$` and each two ascii
characters representing one byte.

It can be used to store opaque data in XML. The DBus end will get a
byte-array.


mtomaschewski (approved):
LGTM

mtomaschewski (approved):

PR #871, opened 2021-05-07
nanny: Fix ni_nanny_identify_node_owner() exit condition

cfconrad:
The function is calling it self recursively with `node->parent`, but
didn't checked if node is already a root one.

This just fix the `Segmentation fault` which was discovered, if we reach
a configuration constraint like:

```xml
<password type="string" constraint="required">
  <meta:user-input type="password" prompt="please enter WPA EAP password"/>
</password>
```
and the password wasn't set in the current configuration.

But it doesn't fix the prompt for password!

### DIFF
--- a/nanny/nanny.c
+++ b/nanny/nanny.c
@@ -529,6 +529,9 @@ ni_nanny_identify_node_owner(ni_nanny_t *mgr, xml_node_t *node, ni_stringbuf_t *
 	ni_managed_device_t *mdev;
 	ni_ifworker_t *w = NULL;
 
+	if (!node)
+		return NULL;
+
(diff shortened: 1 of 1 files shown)


mtomaschewski (approved):
OK, thanks!

We've also to take a look what's going wrong with answering the prompts, but let's defer this until the rewrite to the new wpa-supplicant API is done.

I guess it's the path used to issue the prompt. The `WPA EAP password` is using `"wpa-eap.phase2.password|PASSWORD|%s", essid`, but it is a node inside of a `<wireless><network>` and it is the content of `<wireless>` in the config that is serialized to the method call arguments, so it has to be `network.wpa-eap.phase2.password`.

PR #873, opened 2021-07-01
dracut: fix and improve option handling the cmdline parser (boo#1182227)

mtomaschewski:
- fix to avoid adding interface configs twice
- fix vlan id parsing from vlan interface name in dracut and also in ifcfg
- fix swapped devices, missed lowerdev config relation and improve vlan= (error) option handling
- improve bridge=, bond=, team= option (error) handling, use master relation instead port list (deprecated)
- do not generate config from `ifname=`, but expose the match as pre-parsed parser backend meta option
- refactor to parse parameters in order (e.g. bonding= before ip=)

### DIFF
--- a/client/compat.c
+++ b/client/compat.c
@@ -2934,16 +2934,20 @@ ni_compat_generate_interface(xml_document_t *doc, const ni_compat_netdev_t *comp
 		return NULL;
 
 	if((namenode = xml_node_new("name", ifnode))) {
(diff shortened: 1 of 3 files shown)


Vogtinator:
I can give it a try if you put it somewhere in OBS to build against

mtomaschewski:
> I can give it a try if you put it somewhere in OBS to build against
Yes, I'll merge it into $OBS/network:wicked:testing branch soon.

Issue #878, opened 2021-08-28
Add TUNNEL_GRE_FLAGS support for ALL flags supported by ip-link(8) including ignore-df


archiecobbs:
`TUNNEL_GRE_FLAGS` supports `[i|o]seq` and `[i|o]csum`, but not any other flags that `gre` interfaces can have, such as `ignore-df` (see `ip-link(8)`).

I have an interface `gre1` that needs to have the `ignore-df` flag set, but there's no way to configure this with wicked.

I tried added it to `TUNNEL_GRE_FLAGS` but got this:
```bash
$ wicked ifreload gre1
wicked: ifcfg-gre1: Unsupported TUNNEL_GRE_FLAGS="ignore-df"
wicked: skipping gre1 set-up: no configuration available
```

`TUNNEL_GRE_FLAGS` should support **all** flags supported by `ip-link(8)`, including `ignore-df`.

PR #879, opened 2021-08-28
Add support for the "ignore_df" flag to TUNNEL_GRE_FLAGS (#878).

archiecobbs:
See issue #878 .


mtomaschewski:
Just a wild guess, but commit  c8ce0d1 will cause build issues on SLE-12-SP2 (the oldest supported one).
Further, please split into 2 pull request -- the cleanup has nothing in common with `Add support for the "ignore_df" flag`

archiecobbs:
> the cleanup has nothing in common with `Add support for the "ignore_df" flag`

Not entirely true... some kind of kernel header file update is required because the header files that are there now predate the `ignore_df` flag definition.

But I agree the two changes can be split into two separate PR's which I'll be happy to do. We will end up with a small conflict of course.

archiecobbs:
New separated patch at 791d8376.

mtomaschewski (changes requested):
> > the cleanup has nothing in common with `Add support for the "ignore_df" flag`
> 
> Not entirely true... some kind of kernel header file update is required because the header files that are there now predate the `ignore_df` flag definition.

Yes, to support the ignore-df option on kernels that support it, we need to update the header file.

> But I agree the two changes can be split into two separate PR's which I'll be happy to do.

I'd always also try split the header file update and the actual change to add ignore-df support.

> We will end up with a small conflict of course.

Well, conflicts have to be solved in advance, e.g. by preparing pull#881 branch on top of a master that contains this pull request.
While a cleanup would make sense, I don't think we can remove the header files in pull#881 at all -- see comments there.

But first to this one, tested on sle12-sp2 using kernel 4.4.121-92.152-default first:

- without `ignore_df`:
```
        # grep TUNNEL_GRE_FLAGS ifcfg-gre1
        TUNNEL_GRE_FLAGS="icsum ocsum"
        # ifup gre1
        gre1            up
        # ifdown gre1

        # ip -d monitor link
        7: gre1@NONE: <POINTOPOINT,NOARP> mtu 1472 qdisc noop state DOWN group default
            link/gre 172.20.40.103 peer 172.20.50.1 promiscuity 0
            gre remote 172.20.50.1 local 172.20.40.103 ttl 64 nopmtudisc icsum ocsum

        ==> OK: works
```
- with `ignore_df`:
```
        # grep TUNNEL_GRE_FLAGS ifcfg-gre1
        TUNNEL_GRE_FLAGS="icsum ocsum ignore_df"

        # ifup gre1
        wicked: ifcfg-gre1: Unsupported TUNNEL_GRE_FLAGS="ignore_df"
        wicked: ifup: no matching interfaces

        ==> ifcfg-tunnel.5.in contains a wrong flag name, it has to be` ignore-df`
```
- with proper `ignore-df`:
```
        # grep TUNNEL_GRE_FLAGS ifcfg-gre1
        TUNNEL_GRE_FLAGS="icsum ocsum ignore-df"

        # wicked show-config gre1 | grep ocsum
            <flags>icsum, ocsum, ignore-df</flags>
```
it would be considered.

But `ignore-df` is not a flag (`IFLA_GRE_IFLAGS`, `IFLA_GRE_OFLAGS`, `IFLA_GRE_FLAGS`), but a separate `IFLA_GRE_IGNORE_DF` attribute and needs a separate variable + member in `ni_gre` struct like in the
`IFLA_GRE_OKEY`, `FLA_GRE_PMTUDISC` and other cases.

mtomaschewski:
> I'd always also try split the header file update and the actual change to add ignore-df support.

I mean as separate commits at least (see e.g. commit f5abf5681de78bd98d2036d1dd5614b47bb5ef35).

archiecobbs:
OK sorry - I was hoping this was straightforward but it's beyond my time/ability right now.

PR #881, opened 2021-09-01
Remove legacy Linux headers no longer needed.

archiecobbs:
This PR was split off from PR #878.


mtomaschewski:
It causes to fail already in configure on older distributions (on sle12-sp2):
```
# ./autogen.sh
[...]
checking linux/if_tunnel.h usability... no
checking linux/if_tunnel.h presence... yes
configure: WARNING: linux/if_tunnel.h: present but cannot be compiled
configure: WARNING: linux/if_tunnel.h:     check for missing prerequisite headers?
configure: WARNING: linux/if_tunnel.h: see the Autoconf documentation
configure: WARNING: linux/if_tunnel.h:     section "Present But Cannot Be Compiled"
configure: WARNING: linux/if_tunnel.h: proceeding with the compiler's result
configure: WARNING:     ## --------------------------------------- ##
configure: WARNING:     ## Report this to http://bugs.opensuse.org ##
configure: WARNING:     ## --------------------------------------- ##
checking for linux/if_tunnel.h... no
configure: error: Cannot find required header file
```
the log shows why it fails:
```
In file included from conftest.c:86:0:
/usr/include/linux/if_tunnel.h:31:14: error: 'IFNAMSIZ' undeclared here (not in a function)
  char   name[IFNAMSIZ];
              ^
/usr/include/linux/if_tunnel.h:37:16: error: field 'iph' has incomplete type
  struct iphdr  iph;
                ^
/usr/include/linux/if_tunnel.h:90:19: error: field 'prefix' has incomplete type
  struct in6_addr  prefix;
                   ^
```
so the checks do not work this way.

sle12-sp2 is using `4.4.121-92.152-default` kernel and supports all the tunnels incl. the current flag set
but not the  GRE_IS_SEQ and  similar macros or `IFLA_GRE_IGNORE_DF` from pull#879:
```
        # grep TUNNEL_GRE_FLAGS ifcfg-gre1
        TUNNEL_GRE_FLAGS="icsum ocsum"
        # ifup gre1
        gre1            up

        # ip -d monitor link
        7: gre1@NONE: <POINTOPOINT,NOARP> mtu 1472 qdisc noop state DOWN group default
            link/gre 172.20.40.103 peer 172.20.50.1 promiscuity 0
            gre remote 172.20.50.1 local 172.20.40.103 ttl 64 nopmtudisc icsum ocsum
```

Also, there is still an ethtool-3.x ethtool.h header file which does not support many things (e.g. LINKSETTINGS ioctl).
We require the newer header (ethtool.h, ...) to compile and detect what's supported by the kernel at runtime or catch
with configure options, ifdefs, ....

It's also a quite common thing, that the kernel supports things, that are not supported by the uapi headers provided
in /usr/include/linux/.

Our master is not Tumbleweed only, but as you can see in the `sle*` branches, it is merged + submitted to every
other supported version, potentially incl. into long-time-supported code bases (not equal to package update release)
==> one wicked source for all supported versions.

So we effectively need to update the headers as soon as we add support for these features rather than remove
them completely and break everything except of the most recent version.

archiecobbs:
Well I was afraid of this. I'm a little out of my depth here so I'm going to retract this PR.

However, I would ask a larger question, i.e., whether it is the right strategy to try to handle all possible kernel versions in the same code branch. Would it make more sense to have separate branches for the different kernel/SLE versions?

Just a thought.

PR #882, opened 2021-09-10
ifstatus: Don't show link as "up", if UP and LOWER_UP but not RUNNING

cfconrad:
With wireless we have the case, that the interface is in operstate equal
to DORMANT. This is used to indicate the EAP authentication and set by
the wpa_supplicant. In such state, we have LOWER_UP and UP, but __not__
RUNNING.


mtomaschewski (approved):
Yes, we need an `IFF_LOWER_UP` and `IFF_LOWER_UP` not an `or` to report the link up. Thanks!

Issue #741, opened 2018-03-11
[Question] Multiple ni_process_reap: issue in wickedd logs


MikhailKasimov:
Hello!

Have Leap 42.3 installed on VirtualBox.

Wicked version: 0.6.40.

Multiple ni_process_reap: issue in logs, such like:

```
-- Reboot --
мар 11 18:02:25 linux-mk500 systemd[1]: Starting wicked network management service daemon...
мар 11 18:02:25 linux-mk500 systemd[1]: Started wicked network management service daemon.
мар 11 18:02:28 linux-mk500 wickedd[870]: ni_process_reap: process 1172 has not exited yet; now doing a blocking waitpid()
мар 11 18:13:27 linux-mk500 systemd[1]: Stopping wicked network management service daemon...
мар 11 18:13:27 linux-mk500 systemd[1]: Stopped wicked network management service daemon.
-- Reboot --
мар 11 18:13:46 linux-mk500 systemd[1]: Starting wicked network management service daemon...
мар 11 18:13:46 linux-mk500 systemd[1]: Started wicked network management service daemon.
мар 11 18:13:49 linux-mk500 wickedd[876]: ni_process_reap: process 1176 has not exited yet; now doing a blocking waitpid()
мар 11 19:15:47 linux-mk500 systemd[1]: Stopping wicked network management service daemon...
мар 11 19:15:47 linux-mk500 systemd[1]: Stopped wicked network management service daemon.
-- Reboot --
мар 11 19:17:01 linux-mk500 systemd[1]: Starting wicked network management service daemon...
мар 11 19:17:01 linux-mk500 systemd[1]: Started wicked network management service daemon.
мар 11 19:17:03 linux-mk500 wickedd[892]: ni_process_reap: process 1188 has not exited yet; now doing a blocking waitpid()
```
journalctl -e _PID=1188 command gives no result. So, I can't catch what the problem process (1176, 1188, etc) is.

Can you, please, give a hint what ```ni_process_reap:``` is and how to solve it, if it is problem indeed?

Thanks!
[wickedd_0_6_40_ni_process_reap.txt](https://github.com/openSUSE/wicked/files/1800640/wickedd_0_6_40_ni_process_reap.txt)

archiecobbs:
Still happening 3.5 years later... lots of errors... very verbose...

openSUSE Leap 15.3
wicked-0.6.65-2.1.x86_64

```syslog
$ tail -n 12 /var/log/warn
[...]

PR #883, opened 2021-09-16
process: log command of reaped child (gh#openSUSE/wicked#741)

mtomaschewski:
Changed the error log message to debug with an additional warning message when it took 1sec or more.


cfconrad (approved):
LGTM

PR #884, opened 2021-09-27
Allow a SUSE-style config coexist with a RH-style one.

e4t:
For HPC, warewulf is a popular provisioning tool. It's latest version 4 is a
full reimplementation in Go taking advantage of modern infrastructure and
lessons learned from previous versions.
It provisions a minimal stateless system on a compute node by imaging. The
image is obtained by downloading an OS container from a container registry.
The content of this container is opaque to Warewulf - it doesn't care or know
about the distribution used.
Distribution-specific configuration required by compute nodes is minimal, in
fact limited to the network setup. Warewulf solves this, by adding support for
all 'well known' distributions. Currently, these are RHEL/CentOS/Fedora as
well as Debian & Ubuntu.
While RH-style and Debian-style configurations may coexist, RH-style and
SUSE-style may not: Debian-style systems expect their configuration in the
directory /etc/net while RH-style systems for their configuration in a file
/etc/sysconfig/network - just where SUSE-style distributions (wicked) expect
a directory containing network configuration files.
To remedy this, two possible options exist:
1. Make use of the wicked-native xml configuration - which would
   require to publish this as an external interface.
2. Fix up wicked so it will be able to look for a SUSE-style
   configuration in an alternative location.
This patch implements option 2 (boo#1188442).

Signed-off-by: Egbert Eich <eich@suse.com>

### DIFF
--- a/client/suse/compat-suse.c
+++ b/client/suse/compat-suse.c
@@ -133,6 +133,8 @@ static ni_bool_t		__ni_ipv6_disbled;
 #define __NI_SUSE_PROC_IPV6_DIR			"/proc/sys/net/ipv6"
 
 #define __NI_SUSE_SYSCONFIG_NETWORK_DIR		__NI_SUSE_SYSCONF_DIR"/sysconfig/network"
+// use for OS images supporting both SUSE and RedHat style network configuration
+#define __NI_SUSE_SYSCONFIG_NETWORK_ALT_DIR     __NI_SUSE_SYSCONF_DIR"/sysconfig/network_"
 #define __NI_SUSE_CONFIG_IFPREFIX		"ifcfg-"
 #define __NI_SUSE_CONFIG_GLOBAL			"config"
 #define __NI_SUSE_CONFIG_DHCP			"dhcp"
@@ -220,14 +222,18 @@ __ni_suse_get_ifconfig(const char *root, const char *path, ni_compat_ifconfig_t 
 	char pathbuf[PATH_MAX];
 	char *pathname = NULL;
 	const char *_path = __NI_SUSE_SYSCONFIG_NETWORK_DIR;
+	ni_bool_t default_path = TRUE;
 	unsigned int i;
 
-	if (!ni_string_empty(path))
+	if (!ni_string_empty(path)) {
 		_path = path;
+		default_path = FALSE;
+	}
 
 	if (!root)
 		root = "";
 
+ retry:
 	if (ni_string_empty(root))
 		snprintf(pathbuf, sizeof(pathbuf), "%s", _path);
 	else
@@ -269,7 +275,14 @@ __ni_suse_get_ifconfig(const char *root, const char *path, ni_compat_ifconfig_t 
 						"WAIT_FOR_INTERFACES",
 						&ni_wait_for_interfaces);
 		}
-	} else {
+	} else
+	if (default_path == TRUE) {
+		ni_debug_readwrite("Configuration directory %s is file. Retry with fallback.", pathname);
+		_path = __NI_SUSE_SYSCONFIG_NETWORK_ALT_DIR;
+		default_path = FALSE;
+		goto retry;
+	}
+	else {
 		ni_error("Cannot use '%s' to read suse ifcfg files -- not a directory",
 				pathname);
 		goto done;


cfconrad:
I'm not sure if I understand this change correctly. I will try to summarize what I got so far.

Warewulf is using [system-overlay](https://warewulf.hpcng.org/docs/enchiridion/overlays#system-overlay) to do the pre-configure of a cluster node. Here, warewulf already use the internal wicked-xml confguration interface like [/etc/wicked/ifconfig/ifcfg-eth0.xml.ww](https://github.com/hpcng/warewulf/blob/main/overlays/system/default/etc/wicked/ifconfig/ifcfg-eth0.xml.ww), which contains a basic static IP configuration.

The problem which warewulf trigger is the existence of the **file** [/etc/sysconfig/network](https://github.com/hpcng/warewulf/blob/main/overlays/system/default/etc/sysconfig/network.ww) in the system-overlay, which cause `__ni_suse_get_ifconfig()` to return with a failure [[1]](https://github.com/openSUSE/wicked/blob/master/client/suse/compat-suse.c#L242).

So I agree, it is not possible with "one" system-overlay to cover both SUSE- and RH-style in coexistent, without using xml-files or change the default location of on or the other.

These are my 5 cent to solve this problem
1) Stick to xml-files:
     * Add `/etc/wicked/client.xml` to the `system-overlay` which exclude the usage of `compat:` (same as we see here: https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/wicked/client.xml#L11)
2) Do not throw an error if `/etc/sysconfig/network` exists, but isn't a directory (still uses the `/etc/wicked/ifconfig/ifcfg-eth0.xml` file).
3) Do not use default network path
    * Add `/etc/wicked/client.xml` to the `system-overlay` and specify the suse compat path like 
    ```xml
      <ifconfig location="compat:suse:/etc/sysconfig/network_suse/" />
    ```
   * Advantage, we could still use `ifcfg-*` files.

AFAIK all of these changes are not compatible with other configuration interfaces like yast!

e4t:
Using the internal wicked-xml interface is something Christian has added to warewulf4 to make it work. But we've been informed that this xml representation is not an official API and therefore subject to change without prior notice. Therefore Solution 1. and 2. would be ruled out. Solution 3. is pretty much what my patch did - only that I've used _network as alternative, not network_suse. For the end result it's irrelevant how it's called. The beauty of the name is in the eye of the beholder ;)
For warewulf4 we do not care about YaST at all:
1. on the node you will not run YaST as this is completely ephemeral and thus stateless. If you'd configure this using YaST and reboot, it's gone. The only viable way to configure this is remotely or by placing a configuration into the overlay warewulf uses. Both is outside of the scope of YaST anyway.

cfconrad:
Sure, how we name the "new" network directory doesn't matter :).
And from my perspective placing a `client.xml` into warewulf4's overlay can solve it from wicked perspective.

But what about `sysconfig-netconfig`, it also expect files in `/etc/sysconfig/network/`.

mtomaschewski:
No, there is no **alternative** to the `/etc/sysconfig/network/` directory -- it's not about wicked or the `ifcfg-* files inside, but about **many** other things in the system using it and files/directories like `/etc/sysconfig/network/config` or `/etc/sysconfig/network/scripts/`. When the `/etc/sysconfig/network` path is not a directory but a file, many things including RPM updates of diverse packages including customer scripts will simply fail.

IMO it's much better to remove the `<ifconfig location="compat:"/>` in a warewulf specific `/etc/wicked/client.xml` container config (running the engine) completely than trying to load something "odd" instead. Of course it's not OK to replace the `/etc/sysconfig/network` and disable it or use something else in the deployed SUSE system -- this wouldn't be a SUSE system any more, but be simply unsupported deployment.

About XML: In fact, we need to keep the configuration stable or migrate old config (see https://github.com/openSUSE/wicked/blob/master/src/client/policy.c#L522). A `wicked ifup` parses `ibft`, `ifcfg-*`, `*.xml`  config and sends it to `wickedd-nanny` as xml (enveloped into a policy). Nanny saves it restart-persistently and loads the previously applied config at (re-)start, e.g. while `rcwickedd restart` during RPM updates and executes the `up` flow internally again to "confirm" leases and the like.
It is not "promoted" or "published" as configuation format, because it requires some work & documentation:
 - It's not supported by effectively anything else than wicked on SUSE systems using `ifcfg-*`
 - it's an xml to dbus serialization format defined in the serialization schema where we map nodes in the config to method call arguments and documented in the schema only, there aren't any man pages about.

It's used by Bottlerocked OS in Amazon EKS/ECS or VMWare (see https://github.com/bottlerocket-os/bottlerocket) which is not using ifcfg files and disables `ifcfg-*` support too, what we of course don't want to break too.

More about all this in my comment in the jsc#SLE-21918 feature request, where the discussion belongs.

PR #886, opened 2021-10-24
dbus: config in /usr shouldn't be marked as %config

gmbr3:
%config is only needed when the file is installed to /etc
It shouldn't be used when it is installed to /usr


mtomaschewski (approved):
Thanks!

PR #888, opened 2021-11-10
 ifstatus: Fix warning of unexpected interface flag combination (bsc#1192164)

cfconrad:
Do not log warnings for a valid flag combination on interface state
changes, regression of 9970c28 .


mtomaschewski (approved):
LGTM

PR #892, opened 2021-11-26
wireless: reconnect on unexpected wpa_supplicant restart (bsc#1183495)

cfconrad:
When wpa_supplicant get restarted (e.g. in case of an update) the previous configuration is lost.
As we configure wpa_supplicant via DBus, there is no persistent state by default, thus we need to send the 
configuration again, once wpa_supplicant is back on the DBus.

Bug: https://bugzilla.suse.com/show_bug.cgi?id=1183495


mtomaschewski (approved):
LGTM and worked fine on my notebook, thanks!

PR #894, opened 2021-12-09
cleanup: warnings, time calculations and dhcp fixes (bsc#1188019)

mtomaschewski:
- Cleanup timer/timeout and lifetime types and calculations with very high or
  infinite lease time, which caused high cpu on RaspberryPi 2 (bsc#1188019)
- dhcp6: fix incorrect retransmission time randomization (re-)calculations
  USGv6: DHCPv6_7_1_05c, DHCPv6_7_1_05c, DHCPv6_7_1_05c
- fix warnings from more recent compilers

### DIFF
--- a/src/dhcp6/fsm.c
+++ b/src/dhcp6/fsm.c
@@ -2140,6 +2140,7 @@ ni_dhcp6_fsm_decline(ni_dhcp6_device_t *dev)
 
 
 	if (dev->retrans.count == 0) {
+		ni_dhcp6_fsm_timer_cancel(dev);
 
 		if (!ni_dhcp6_fsm_decline_info(dev, dev->lease->dhcp6.ia_list,
 				"Initiating DHCPv6 lease addresses decline",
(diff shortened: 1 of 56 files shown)


src/util.c
```diff
@@ -1445,16 +1445,16 @@ ni_bool_t
...
+	int ret;
 
 	if (!fmt)
 		return FALSE;
 
 	va_start(ap, fmt);
 	ret = vasprintf(&path, fmt, ap) > 0;
 	va_end(ap);
-	if (!ret)
+	if (ret < 0 || !path)
```
cfconrad, line 1457: maybe it's time to switch to `ni_string_printf()` instead...

src/dhcp6/fsm.c
```diff
@@ -2641,8 +2641,9 @@ ni_dhcp6_fsm_ia_addr_delete(ni_netdev_t *ifp, ni_dhcp6_device_t *dev, const ni_a
 		}
 	}
 
-	if (duplicate)
+	if (duplicate) {
```
cfconrad, line 2644: If I not mistaken, our code style is to avoid brackets on single statements

src/util.c
mtomaschewski: This wouldn't work but require a `ni_string_vprintf` not `ni_string_printf`. The intention was to fix a warning/bug, not to refactor the functions.

src/dhcp6/fsm.c
mtomaschewski: yes.

src/util.c
cfconrad: ack

cfconrad (approved):
LGTM, thx!

PR #896, opened 2021-12-16
Fix sysctl file read precedence

cfconrad:
* Respect sort order of filenames from sysctl.d directories (complete commit 140a945645a1292d4bb5efd240f02f386ae70ae6).
 * Display sysctl key value pairs in debug
 * Add `ni_var_array_sort()` function.


mtomaschewski (approved):
LGTM

PR #899, opened 2022-02-14
gitlab-ci: Use CI_PROJECT_NAMESPACE for wicked-ci

cfconrad:
Use the wicked-ci.yml from the same `PROJECT_NAMESPACE` as the wicked
repo belongs to. E.g. if the CI is running on the project `cfconrad/wicked`
it will also use the `wicked-ci.yml` from `cfconrad/wicked-ci`.


mtomaschewski (approved):
LGTM, thanks!

PR #900, opened 2022-02-21
compat-suse: Fix reading of sysctl varialbe addr_gen_mode

cfconrad:
Previously the value of sysctl addr_gen_mode was internally set to
the value of use_tempaddr.


mtomaschewski (approved):
LGTM, thanks!

PR #901, opened 2022-03-10
Fix: Apply DNS from RA rdnss after interface ifdown/ifup (bsc#1181429)

cfconrad:
On ifdown, the call `ni_auto6_release()` set the `auto6->enabled` state
to FALSE. This state is never changed to TRUE even if we do a explicit
`ni_auto6_acquire()`.


mtomaschewski (approved):
LGTM, thanks!

PR #903, opened 2022-03-15
buffer: adjust uint length to size_t, guard init

mtomaschewski:
- adjust unsigned int buffer count and tailroom return type to size_t
- add guards and return status in init and ensure_tailroom functions
- adopt the using side to use size_t and zu printf formats


cfconrad (approved):
LGTM, thx

PR #905, opened 2022-03-22
compat-suse: Match sysctl.d read order with systemd-sysctl

cfconrad:
Match the read order of sysctl.d directories like systemd-sysctl.
progps-ng does this since 3.3.17 [1].

[1] https://gitlab.com/procps-ng/procps/-/commit/5da3024e4e4231561d922ad356a22c0d5d7bc69f


mtomaschewski (approved):
LGTM, thanks!

PR #907, opened 2022-03-31
xml-schema: Fix refcount handling for ni_xs_type_t

cfconrad:
If the loaded xml-schema is errournous (e.g. a typo or missing include),
the cleanup can hit a ni_assert(), because of missing refcount increment
of the "array-element" type referenced in "array" type.


mtomaschewski (approved):
Tested, works fine for me.

PR #908, opened 2022-03-31
wireless: Fix WPA_PSK given as 64byte long hex string

cfconrad:


mtomaschewski (approved):
OK, it wrongly expected it to be `xx:...:xx` colon separator in the psk hash string.

Issue #910, opened 2022-04-17
wireless-tools / libiw is dead, please replace with libnl3


Conan-Kudo:
With Fedora Linux 36, `wireless-tools` has been dropped due to lack of upstream support and being superseded by Linux kernel native UAPIs for over a decade. These APIs are generally available from the `libnl3` library, which Wicked already depends on.

However, Wicked still depends on `wireless.h` from `wireless-tools-devel` (`libiw-devel` in openSUSE). Please migrate to `libnl3` for these remaining cases.

Information for `libnl3` is at: http://www.infradead.org/~tgr/libnl/

Reference: https://fedoraproject.org/wiki/Changes/RemoveWirelessExtensions

cfconrad:
release 0.6.70 should get it

PR #872, opened 2021-05-31
wireless: migrate to wpa-supplicant v1 DBus interface and improve

cfconrad:
Change the wicked wireless module to work with latest wpa_supplicant DBus interface https://w1.fi/wpa_supplicant/devel/dbus.html

Addressed issues with this PR:
 * wpa_supplicant-2.9 is broken with wicked [bsc#1156920](https://bugzilla.opensuse.org/show_bug.cgi?id=1156920)
 * Cannot connect to EAP secured wireless network with wicked [bsc#1057592](https://bugzilla.suse.com/show_bug.cgi?id=1057592)
   * TLS, TTLS, MSCHAPv2 tested
 * Show wireless connection status and scan-results via `wicked`
   * The half of [bsc#1160654](https://bugzilla.suse.com/show_bug.cgi?id=1160654)
 * Allow `scan-only` wireless configurations
 * Allow multiple wireless networks configurations per interface
 * Adopt `man ifcfg-wireless` to latest changes
 * Read `WIRELESS_REGULATORY_DOMAIN` from `/etc/sysconfig/network/config`
 * Fix WEP connection

PR #911, opened 2022-04-25
wireless: remove libiw dependency

cfconrad:
Remove libiw dependency (since https://github.com/openSUSE/wicked/pull/872)
used to detect wireless interfaces only.
Unfortunately `rtnetlink(7)` does not provide this info in the `NEWLINK`
message, so we're using sysfs telling us if an interface has a wireless
extension, once the `/sys/class/net/$name/wireless` directory exists
(issue: https://github.com/openSUSE/wicked/issues/910).

### DIFF
--- a/src/iflist.c
+++ b/src/iflist.c
@@ -1138,12 +1138,9 @@ __ni_process_ifinfomsg_linktype(ni_linkinfo_t *link, const char *ifname, ni_netc
 			/* We're at the very least an ethernet. */
 			tmp_link_type = NI_IFTYPE_ETHERNET;
 
-			/*
-			 * Try to detect if this is a  WLAN device.
-			 * The official way of doing this is to check whether
-			 * ioctl(SIOCGIWNAME) succeeds.
-			 */
-			if (__ni_wireless_get_name(ifname, NULL, 0) == 0)
+			/* rtnetlink does not tell us if the device has a
+			 * wireless extensions or not, but sysfs does. */
+			if ( ni_sysfs_netif_exists(ifname, "wireless"))
 				tmp_link_type = NI_IFTYPE_WIRELESS;
 
 			memset(&drv_info, 0, sizeof(drv_info));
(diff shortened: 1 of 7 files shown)


src/iflist.c
```diff
@@ -1138,12 +1138,9 @@ __ni_process_ifinfomsg_linktype(ni_linkinfo_t *link, const char *ifname, ni_netc
 			/* We're at the very least an ethernet. */
 			tmp_link_type = NI_IFTYPE_ETHERNET;
 
-			/*
-			 * Try to detect if this is a  WLAN device.
-			 * The official way of doing this is to check whether
-			 * ioctl(SIOCGIWNAME) succeeds.
-			 */
-			if (__ni_wireless_get_name(ifname, NULL, 0) == 0)
+			/*  Netlink does not tell us if the device has a
```
mtomaschewski, line 1141: Please use `rtnetlink` unstead of (unspecified)netlink:
```
/*  rtnetlink does not tell us if the device has a
*  wireless extenstions or not, but sysfs does. */
```
Also `extention` is incorrect spelling of `extension`.

mtomaschewski (changes requested):
```
It was only used to determine if a interface name belongs to a
wireless interface or not.
Unfortunately netlink do not provide this info in the NEWLINK message,
But sysfs tell us if a interface has a wireless extension, once a
/sys/class/net//wireless directory exists.
```

Please use more explicit `rtnetlink` instead just netlink. Also `do` -> `does`, `tell` -> `tells`, `But` -> `but` (or end the sententce and start new one).

I'd suggest:

Remove `libiw` dependency (since https://github.com/openSUSE/wicked/pull/872)
used to detect wireless interfaces only.  Unfortunately `rtnetlink(7)` does not
provide this info in the `NEWLINK` message, so we're using sysfs telling us if
an interface has a wireless extension, once the `/sys/class/net/$name/wireless`
directory exists (issue https://github.com/openSUSE/wicked/issues/910).

cfconrad:
requested changed applied

mtomaschewski (approved):
Thanks, LGTM.

PR #912, opened 2022-05-04
ifevent: always close old rtnl event sock on restart

cfconrad:
The used `ni_socket_release()`, while restart on rtnetlink event
errors, didn't closed and deactivated/removed the obsolete socket
from poll processing immediately. This caused further restarts and
finally an assertion due to incorrect reference count (bsc#1192508).


mtomaschewski (approved):
LGTM, we've tested this extensively already.

PR #915, opened 2022-05-20
client: add --[no-]release option to ifdown/ifreload (jsc#SLE-10249)

mtomaschewski:
Add `wicked <ifdown|ifreload> --[no-]release ..` option, that overrides the configured
release behavior previously applied by ifup and force to (not) release dhcp leases.


cfconrad (approved):
LGTM, Thx!

PR #917, opened 2022-06-09
util: Add bitmask utility functions

cfconrad:


src/util.c
```diff
@@ -37,6 +37,9 @@
 #define NI_BYTE_ARRAY_CHUNK	0xff	/* important: (2^n)-1 with all bits set */
 #define NI_BYTE_ARRAY_SIZE(len)	((len) | NI_BYTE_ARRAY_CHUNK)
 
+#define NI_BIT_SEPERATOR_FORMAT	"|"
```
mtomaschewski, line 40: The parsing of the bitmask (names for values) and bitmap (names for bit numbers) types is the same, but they're differently formatted:
```
ni_dbus_deserialize_xml_scalar
        if (scalar_info->constraint.bitmask) 
                format: " | "
        if (scalar_info->constraint.bitmap)
                format: ", "
```
Perhaps we should  use different default separators for both:
```suggestion
#define NI_BITMAP_SEPERATOR_FORMAT	","
#define NI_BITMASK_SEPERATOR_FORMAT	"|"
```

src/util.c
```diff
@@ -37,6 +37,9 @@
...
+#define NI_BIT_SEPERATOR_PARSE	" ,|\t\n"
```
mtomaschewski, line 41: Also for parsing, even it's the same?

mtomaschewski (commented):
LGTM, but I'd use different `FORMAT` constants for both.

src/util.c
cfconrad: Ack I will change it!

src/util.c
cfconrad: For parsing we allow almost all separators, see
* https://github.com/openSUSE/wicked/blob/master/src/dbus-xml.c#L563
* https://github.com/openSUSE/wicked/blob/master/src/dbus-xml.c#L600

But maybe we could do it for consistency.

mtomaschewski: Yes, the parsing of both is the same.

cfconrad:
This PR change the default separator for `ni_format_bitmap()` and `ni_format_bitmap_string()` from `|` to `,`.

Breaking changes:
1) `ni_ethtool_wol_flags_format()`
     The only caller is in `ni_ethtool_set_wake_on_lan()` and it appear only in debug/warn messages.
2) `src/names.c`
   * `ni_addrconf_flags_format()` => no call without specify the separator 
   * `ni_addrconf_update_flags_format` => no call without specify the separator
   * `ni_dhcp6_mode_format()`
     * client/compat.c store `<ipv6:dhcp><mode>XXX</mode>/<ipv6:dhcp>`
     * src/dhcp6/fsm.c only in info messages
     * src/dhcp6/device.c only in debug messages

src/util.c
cfconrad: done

cfconrad:
Ready from my side

mtomaschewski (approved):
LGTM

PR #919, opened 2022-06-24
xpath-fmt: Fix segfault when formating an empty node

cfconrad:
If we try to format an empty node with xpath, we got a segfault
in the corresponding printf call, e.g.
```
echo "<foo><bar></bar><foo>" | \
  wicked xpath --reference '/foo' '%{bar}'
```


mtomaschewski (approved):
LGTM

PR #922, opened 2022-07-06
wireless: minimize extension/type detection race on ifrename

cfconrad:
If we process the ifinfomsg_linktype, but the interface was renamed
in the meantime, the corresponding sysfs path doesn't exists
anymore. Minimize the race possibility by requesting the new ifname.


mtomaschewski (approved):
Thanks!

PR #923, opened 2022-07-07
wireless: <auth-proto> prefer wpa/rsn values over wpa1/wpa2

cfconrad:
The variable `WIRELESS_WPA_PROTO`, the field `<auth-proto>` and the capability
field `<wpa-protocols>` referencing all the same bitmap. And we would
like to see wpa/rsn instead of wpa1/wpa2.
For backward compatibility wpa1/wpa2 can still be used but
is not generated on converting ifcfg to XML.


mtomaschewski (changes requested):
Please use `prefer`, not `prefere` in commit logs.

Does a config with an explicit scan match not need also explicit proto + ciphers settings (any more)? Would be better to make it in a separate commit rather than as "unrelated" change.

cfconrad:
As discussed, we do not include the change of the `WIRELESS_AP_SCANMODE` cipher selection within this PR!

mtomaschewski (approved):
OK, thanks.  We can change the current behavior + document or whatever in separate pull.

PR #926, opened 2022-07-08
addrconf: remove address before calling (dhcp) release

mtomaschewski:
USGv6: DHCPv6_1_2_07b, According to RFC 3315 section 18.1.6, ”The client
MUST NOT use any of the addresses it is releasing as the source address
in the Release message or in any subsequently transmitted message.”

We didn't used the dhcp address as source in release, but because the
address were not yet removed from the interface, the host were still
able to use it and responded to echo request (ping) during the release,
causing an USGv6 testsuite failure.


src/dbus-objects/addrconf.c
```diff
@@ -636,80 +614,348 @@ ni_objectmodel_addrconf_forward_request(ni_dbus_addrconf_forwarder_t *forwarder,
...
+		return -1;
+
+	ni_debug_objectmodel("%s: forwarding drop for lease=%s:%s in state=%s uuid=%s",
+			dev->name,
+			ni_addrfamily_type_to_name(lease->family),
+			ni_addrconf_type_to_name(lease->type),
+			ni_addrconf_state_to_name(lease->state),
+			ni_uuid_is_null(&lease->uuid) ? "null" : ni_uuid_print(&lease->uuid));
+
+	if (ni_addrconf_drop_args_into_variant(action->fwdr, &var, &action->args))
```
cfconrad, line 707: I think we miss a `ni_dbus_variant_free(&var)` at the end.

src/dbus-objects/addrconf.c
```diff
@@ -636,80 +614,348 @@ ni_objectmodel_addrconf_forward_request(ni_dbus_addrconf_forwarder_t *forwarder,
...
+static int				ni_addrconf_drop_collect_exec(ni_netdev_t *, ni_addrconf_lease_t *);
+static ni_addrconf_action_t *		ni_addrconf_drop_collect_complete(ni_addrconf_lease_t *);
+
+static inline void
+ni_addrconf_drop_args_init(const ni_dbus_addrconf_forwarder_t *fwdr, ni_addrconf_drop_args_t *args)
 {
-	ni_uuid_t uuid = NI_UUID_INIT;
-	ni_addrconf_lease_t *lease;
-	int rv = -1;
+	(void)fwdr;
```
cfconrad, line 642: don't we have a unused macro?

src/dbus-objects/addrconf.c
```diff
@@ -636,80 +614,348 @@ ni_objectmodel_addrconf_forward_request(ni_dbus_addrconf_forwarder_t *forwarder,
...
+
+	if (pos && ni_addrconf_action_list_insert(&pos->next, &forward->action)) {
+		if (ni_addrconf_action_list_append(&pos->next, &collect->action))
+			return TRUE;
+	}
+
+failure:
+	ni_addrconf_drop_collect_free(collect);
+	ni_addrconf_drop_forward_free(forward);
+	return FALSE;
```
cfconrad, line 860: This isn't 100% save, if `if (ni_addrconf_action_list_append(&pos->next, &collect->action))` fails we free `forward` which is already in the list.
On the other hand, there is no way to fail...

src/dbus-objects/addrconf.c
```diff
@@ -636,80 +614,348 @@ ni_objectmodel_addrconf_forward_request(ni_dbus_addrconf_forwarder_t *forwarder,
...
+	ni_addrconf_drop_collect_t *collect;
+	ni_addrconf_action_t *action;
+
+	if (!lease || !lease->updater)
+		return NULL;
+
+	action = ni_addrconf_action_list_find_exec(lease->updater->action,
+					ni_addrconf_drop_collect_exec);
+	if (!(collect = (ni_addrconf_drop_collect_t *)action))
+		return NULL;
```
cfconrad, line 875: Just a question:
This code expect only one of the `ni_addrconf_drop_collect_t*` objects inside the action list. But I do not see a guard when we adding this type to the list. Do we need to guard it?

cfconrad (changes requested):
Changes of 5359abfd and c0734e65 do not compile, if I remember correctly we don't wanna have such commits, right?

src/dbus-objects/addrconf.c
mtomaschewski: We actually don't need the fwdr to decide how to init it -- I've added the parameter for consistency, but we can remove it completely too.

src/dbus-objects/addrconf.c
mtomaschewski: ni_addrconf_action_list_insert or append cannot fail or in another words, they can fail only when pos would be NULL (what we catch) or when we've failed to allocate forward or collect (both catched before too). We can  do this  instead: 
```diff
@@ -849,9 +850,11 @@ ni_addrconf_drop_actions_arm(ni_dbus_addrconf_forwarder_t *forwarder,
        pos = ni_addrconf_action_list_find_exec(updater->action,
                                ni_addrconf_action_addrs_remove);
 
-       if (pos && ni_addrconf_action_list_insert(&pos->next, &forward->action)) {
-               if (ni_addrconf_action_list_append(&pos->next, &collect->action))
-                       return TRUE;
+       if (pos) {
+               /* insert or append cannot fail -- checked conditions above */
+               ni_addrconf_action_list_insert(&pos->next, &forward->action);
+               ni_addrconf_action_list_append(&pos->next, &collect->action);
+               return TRUE;
        }
 
 failure:
````

File: src/dbus-objects/addrconf.c (hunk 4 above)
Comment (mtomaschewski) [reply src/dbus-objects/addrconf.c:875]: The guard is the `ni_addrconf_drop_collect_exec` -- we search for the collect action using it. When we don't find it, `action == NULL` (and also the cast to collect). The above code could be also:
```C
action = ni_addrconf_action_list_find_exec(lease->updater->action, ni_addrconf_drop_collect_exec);
if (!action)
	return NULL;
collect = (ni_addrconf_drop_collect_t *)action;
```
It's just written in one line.

src/dbus-objects/addrconf.c
mtomaschewski: Should I change to the above variant to avoid confusion about?

mtomaschewski:
> Changes of [5359abf](https://github.com/openSUSE/wicked/commit/5359abfd031ebcb950f1513a60a2c1c883d3d9a6) and [c0734e6](https://github.com/openSUSE/wicked/commit/c0734e6518c6c4f93f31c4f40eac78488b0020b8) do not compile, if I remember correctly we don't wanna have such commits, right?

I've checked them out using:
```
make maintainer-clean ; git checkout -b test-c0734e6 c0734e6 && ./autogen.sh &&  make -s  V=0 ; git status
make maintainer-clean ; git checkout -b test-5359abf 5359abf && ./autogen.sh &&  make -s  V=0 ; git status
make maintainer-clean ; git checkout -b test-49adbbb 49adbbb && ./autogen.sh &&  make -s  V=0 ; git status
```
and don't see any issue / warning / ... Can't reproduce, they compile fine for me.

src/dbus-objects/addrconf.c
mtomaschewski: removed the unused param in https://github.com/openSUSE/wicked/compare/41a0903..0c9aacc

src/dbus-objects/addrconf.c
mtomaschewski: Adjusted to avoid confusion in https://github.com/openSUSE/wicked/compare/41a0903..0c9aacc

cfconrad:
> and don't see any issue / warning / ... Can't reproduce, they compile fine for me, also `make -s V=0 distcheck rpmbuild` works.

Confirmed, it was my environment which wasn't clean, sorry for the circumstances !

src/dbus-objects/addrconf.c
cfconrad: I wonder why you added it twice, do I oversee something?

src/dbus-objects/addrconf.c
mtomaschewski: > expect only one of the ni_addrconf_drop_collect_t* objects inside the action list

Ahh.. now I understand, you mean there are multiple collect actions in  the action list like:
   ```updater->actions->[collect]->..->[collect]->..```
and we complete only the 1st one, so the  2nd would continue to collect/wait.

No, this can't happen. We always allocate new releasing lease and a new updater with a fresh action list where we add  only 1 forward action behind `remove address` (see arm funct) and 1 collect action as last action in the list.

cfconrad (approved):
LGTM

PR #928, opened 2022-07-22
compat-team: Set <link> node in slaves (bsc#1200505)

cfconrad:
The xml configuration of the slave missed
the `<link>` configuration, when generated from
ifcfg format.


mtomaschewski (approved):
LGTM

PR #931, opened 2022-09-29
wpa_supplicant: Fix crash on unexpected restarts of wpa_supplicant

cfconrad:
Once wpa_supplicant restart unexpected, wickedd tend to crash, if the wpa_supplicant handler code was currently in one of the
DBus calls. To fix this, we need refcounting for the DBus objects of wpa_supplicant and be aware of such a situation in error handing.

### DIFF
--- a/src/wireless.c
+++ b/src/wireless.c
@@ -445,6 +445,7 @@ __ni_wireless_scan_timeout(void *ptr, const ni_timer_t *timer)
 
 	ni_wireless_trigger_scan(dev, wif, FALSE);
 	__ni_wireless_scan_timer_arm(scan, dev, scan->interval);
+	ni_wpa_nif_drop(&wif);
 }
 
 static void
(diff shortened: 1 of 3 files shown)


mtomaschewski (changes requested):
LGTM, but please fix indenting in the commit message:
```
commit 3309697d1dd9eee6470a04c63f84e630695810e9
Author: Clemens Famulla-Conrad <cfamullaconrad@suse.de>
Date:   Thu Jul 7 22:18:08 2022 +0200

    wpa-supplicant: Add refcounting for ni_wpa_bss_t
    
    Each call on DBus can result in async receving of DBus signals.                                                                                                                                                                         Thus we need to protect our current working object, that it                                                                                                                                                                             do not get removed because of such a signal.
```

cfconrad:
Changed the commit message and rebased on top of master.

mtomaschewski (approved):
LGTM

PR #935, opened 2022-10-17
testing: Null terminate map in bitmap-test

jcronenberg:
Maps are supposed to end with a NULL element as otherwise out of bounds reads can occur.


cfconrad (approved):
LGTM, thx!

PR #942, opened 2022-12-21
ifstatus: improve to show a no-carrier status

mtomaschewski:
Split the transient `setup-in-progress` status and show `no-carrier` status
when the interface is set administratively up (device-up), but the (L2)
link is not up and running yet (link-up), including e.g. not-associated
wireless links. The following (L3) setup like IP / lease acquisition
continues to appear as setup-in-progress until the setup is sufficient
and we consider the interface it as up.


cfconrad (approved):
LGTM, thx

PR #945, opened 2023-01-24
nanny: fix use-after-free in debug mode (bsc#1206447)

mtomaschewski:
Fix a use-after-free by getting a local mpolicy->fsm_policy reference in unregister_managed_policy to safely access the fsm policy name for a debug message logged after unregistering the mpolicy object, which is freeing the mpolicy (if it's the last reference).

A policy deletion is performed by wicked ifdown or ifreload and the execution is limited to root (uid 0) callers, so it does not bring additional privilege possibilities.


cfconrad (approved):
LGTM

PR #947, opened 2023-01-25
client: add firmware enable|disable command (jsc#PED-3132)

mtomaschewski:
* Improved parsing of the `/etc/wicked/*.xml` config files to handle
  different extension types and overrides of their script and builtin
  action definitions.
* Implemented a `wicked firmware <enable|disable> <name>` command to
  enable or disable netif-firmware-discovery scripts by generating a
  `client-firmware.xml` override config file.
* Adopted `wicked redfish <enable|disable>` to use `wicked firmware`.


cfconrad (approved):
LGTM

PR #952, opened 2023-01-30
bond: workaround 6.1 enslave regression (boo#1206674)

mtomaschewski:
The kernel bonding accepts only enslave of interfaces in DOWN state and the bonding driver sets the slave UP itself after enslave, once it also inherited MTU, MAC address, ... to the slave as needed.

All kernels between (at least) 4.12 up to 6.0 accepted to enslave using a single `ip link set down master bond0 dev eth0` rtnetlink message, which ensured that the (eth0) interface to enslave were set down if needed and wasn't rejected to enslave.

Since kernel 6.1 (commit a4abfa627c3865c37e036bccb681619a50d3d93c) the enslave processing changed and the `down` flag in the enslave message is applied after enslave, causing that the UP flag added by enslave is removed again, the slave remains inactive and is not used by the bond, which remains in a no-carrier state.

Changed to not pass the DOWN flag any more, but enslave only.

### DIFF
--- a/src/ifconfig.c
+++ b/src/ifconfig.c
@@ -127,7 +127,7 @@ static int	__ni_rtnl_link_unenslave(const ni_netdev_t *);
 static int	__ni_rtnl_link_delete(const ni_netdev_t *);
 
 static int	__ni_rtnl_link_add_port_up(const ni_netdev_t *, const char *, unsigned int);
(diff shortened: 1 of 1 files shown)


mtomaschewski:
Tweaked the commit message a bit.

cfconrad (approved):
THX

PR #954, opened 2023-02-14
testing: port json-test to wunit framework and extend it

jcronenberg:


cfconrad:
We are using these `<type>: ` prefix for the commit messages. I think we should use `json: ` for 04f834a and `testing: ` for d43e658.

cfconrad (approved):
Thx, LGTM!

PR #957, opened 2023-02-27
refcount: split declaration and definition macros

mtomaschewski:
Changed to use ni_declare|ni_define as prefix
and split declaration and definition macros.


cfconrad (approved):
LGTM, thx!

PR #962, opened 2023-04-20
client: improve netif-firmware-discovery extension

mtomaschewski:
- appconfig: refactor firmware extension definition to use one 
  extension per firmware as in system-updater and dbus-service 
  and consider environment variables specified in the extension 
  and script action scopes. See also `wicked-config(5)`.
- client: fix enable|disable to add only and override config to
  the client-firmware.xml for requested firmware without to 
  repeat details like script action command (causing issues).
  Add a revert command to revert changes by enable|disable and
  options to choose xml vs text table format.
- man: add `wicked-firmware(8)`, update `wicked-config(5)`
- spec: cleanup nbft config on wicked-nbft uninstall


cfconrad (approved):
LGTM

PR #965, opened 2023-05-08
autogen.sh: improve usability and enable silent rules

mtomaschewski:
- Add `-Werror` option to fail on compiler warnings and `-debug0` for non-optimized
  build (`-g -O0` plus) for an intended gdb use.
- Enable silent-rules to improve warnings/issues visibility in favor of full libtool output
  and pass all other options to configure additionally and not instead of our defaults,
  now considering also _rundir rpm config.


cfconrad (approved):
THX, looks good to me!

PR #966, opened 2023-05-10
wicked-nbft: conditionally disable on SLE < 15-SP5

mtomaschewski:
Add rpmbuild --with[out] nbft conditional along with the configure --disable-nbft option
to not build the wicked-nbft firmware extension rpm sub-package on SLE/LEAP < 15.5.


cfconrad (approved):
LGTM, thx

PR #977, opened 2023-06-14
testing: run nbft-test.sh only if jq is available

mtomaschewski:
Execute nbft-test.sh in make check only when nbft
is enabled and also the jq utility is available.


cfconrad (approved):
LGTM thx!

PR #978, opened 2023-06-14
array: add macro to define qsort cmp function

mtomaschewski:
Add macro to define qsort element compare wrapper in favour
of a nested wrapper inside the array qsort function to avoid
executable stack requirements.


cfconrad (approved):
LGTM, Thx

PR #979, opened 2023-06-15
 arp: Improve error handling and default settings 

cfconrad:
* Read probe and notify settings from config file
* Retry arp verify/notify on ENOBUFS
* Unify arp verify/notify in autoip4, dhcp4 and static-ip
* client: allow interval as range for `wicked arp verify`


mtomaschewski (approved):
LGTM, thanks!

PR #982, opened 2023-07-03
ifconfig: fix arp notify loop (boo#1212806) and burst sending

cfconrad:
* fix the regression [boo#1212806](https://bugzilla.suse.com/show_bug.cgi?id=1212806) introduced with https://github.com/openSUSE/wicked/pull/979 
 * fix the burst sending of max 256 arp verify/notify messages
 * config value of `<arp><verify|notify><count>` need to be greater then `0`


mtomaschewski (approved):
OK, I've retested the cases I've had in my mind and it LGTM. Thanks!

PR #985, opened 2023-10-09
system-updater: Parse updater format from XML configuration to ensure install calls can run.

boomer41:
In ni_system_updaters_init(), the updater's format is set by using ni_updater_format_type(ex->format). This in turn is used by ni_system_updater_generic_leaseinfo_create(), which is called by ni_system_updater_generic_install_call(). As ex->format is never actually set, ni_system_updater_generic_leaseinfo_create will always fail with an unknown format error and the install call is never executed.

### DIFF
--- a/src/appconfig.c
+++ b/src/appconfig.c
@@ -2025,6 +2025,7 @@ ni_config_parse_system_updater(ni_extension_t **list, xml_node_t *node)
 {
 	ni_extension_t *ex, *old;
 	const char *name;
+	const char *format;
 
 	ni_assert(list && node);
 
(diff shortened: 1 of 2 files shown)


src/appconfig.c
```diff
@@ -2025,6 +2025,7 @@ ni_config_parse_system_updater(ni_extension_t **list, xml_node_t *node)
 {
 	ni_extension_t *ex, *old;
 	const char *name;
+    const char *format;
```
mtomaschewski, line 2028:
```suggestion
	const char *format;
```

mtomaschewski (changes requested):
Please fix the indenting.

mtomaschewski:
This is obviously a **regression** in **version-0.6.71..version-0.6.72**.

src/appconfig.c
boomer41: Sorry, forgot to change that one to tabs. Fixed with new commit.

mtomaschewski (approved):
Thanks!!

boomer41:
> This is obviously a **regression** in **version-0.6.71..version-0.6.72**.

This is our observation as well.
v0.6.69 worked, and the current version for openSUSE 15.4 (v0.6.72) doesn't work.

mtomaschewski:
> This is our observation as well. v0.6.69 worked, and the current version for openSUSE 15.4 (v0.6.73) doesn't work.

I assume, you're using a custom updater instead of netconfig as netconfig is not using `install`+`remove` actions,
but `batch` (already SLE-12-SP2 has it).

boomer41:
> I assume, you're using a custom updater instead of netconfig as netconfig is not using `install`+`remove` actions, but `batch` (already SLE-12-SP2 has it).

Yes, we're running our custom scripts using install/remove actions, which weren't called anymore after the update.

PR #989, opened 2023-11-01
duid: fix comment for v6time

bytesbymike:
v6time is seconds since midnight (UTC), January 1, 2000, modulo 2^32. The value is generated correctly but the comment refers to 2001 instead of 2000.


mtomaschewski (approved):
Thanks!

PR #994, opened 2024-01-25
client: fix exit code for no-carrier status (bsc#1219265)

coderliss:
The 'no-carrier' status was separated from the 'setup-in-progress' status. Even though the interface appears active during this status, the setup process is not complete and the NIC is not ready. Therefore, the `ifstatus <nic>` command should not return 0 for this status.

### DIFF
--- a/client/ifstatus.c
+++ b/client/ifstatus.c
@@ -884,6 +884,7 @@ ni_do_ifstatus(int argc, char **argv)
 			case NI_WICKED_ST_UNCONFIGURED:
 			case NI_WICKED_ST_NOT_RUNNING:
 			case NI_WICKED_ST_IN_PROGRESS:
+			case NI_WICKED_ST_NO_CARRIER:
 				break;
 			default:
 				status = NI_WICKED_ST_OK;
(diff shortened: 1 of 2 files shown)


coderliss:
My application operators the NIC card and use `ifstatus <nic>` to check the interface is ready or not. and then retrieve the IP addresses from the NIC if it is ready. but it occasionally fails to get IP after upgrading to version 0.6.73.

I debug the issue by replacing the command with `wicked ifstatus –transient` and discovered that sometimes exit code 168 (NI_WICKED_ST_NO_CARRIER) returned.

I guess the issue is related to the commit ced5db71c83e3ffec6a14feac0f7d28f6f06cfef
Please help review this PR, Thanks.

cfconrad:
Good catch, thx. From my point this looks good.
I would add the same case in 
https://github.com/openSUSE/wicked/blob/744ed4160483e5b78d60b7c0f86ae0c37802cd22/client/ifstatus.c#L1003
so we get also consistent behavior in `ifup` and `ifreload`.

cfconrad (approved):
LGTM

coderliss:
Thank you for review!

PR #998, opened 2024-02-19
vxlan: don't format unknown rtnl attrs (bsc#1219751)

mtomaschewski:


schema/vxlan.xml
```diff
@@ -52,7 +52,7 @@
 
  <method name="changeDevice">
   <arguments>
-   <config type="vxlan:base-config">
```
cfconrad, line 55: What is the purpose of `base-config` then?
I think we could simply add the `address` field to `configuration`?

mtomaschewski: It's OK as  it is.

The `configuration` dict contains everything that vxlan interface supports, but `address` (`IFLA_ADDRESS` aka MAC) is not vxlan  specific  (`IFLA_VXLAN*`), but ethernet-type specific.

PR #999, opened 2024-02-20
xpath: allow underscore in node identifier

cfconrad:
We use underscores in our own XML schema. So lets allow such names in `wicked xpath --reference XXX`.


mtomaschewski (changes requested):
As you  touch it,  please use just
```
 *      You should have received a copy of the GNU General Public License
 *      along with this program. If not, see <http://www.gnu.org/licenses/>.
```
instead.

mtomaschewski (approved):
OK, thanks!

PR #1002, opened 2024-02-23
version-0.6.74

cfconrad:
- team: add new options like link_watch_policy (jsc#PED-7183)
- Fix memory leaks in dbus variant destroy and fsm free
- xpath: allow underscore in node identifier
- vxlan: don't format unknown rtnl attrs (bsc#1219751)
- ifreload: VLAN changes require device deletion (bsc#1218927)
- ifcheck: fix config changed check (bsc#1218926)
- client: fix exit code for no-carrier status (bsc#1219265)
- dhcp6: omit the SO_REUSEPORT option (bsc#1215692)
- rtnl: parse peer address on non-ptp interfaces
- system-updater: Parse updater format from XML configuration to ensure install calls can run.
- ifconfig: fix arp notify loop (boo#1212806) and burst sending


mtomaschewski (approved):
LGTM

PR #1003, opened 2024-03-14
addrconf: fix fallback-lease drop (bsc#1220996)

cfconrad:
This PR fix [bsc#1220996](https://bugzilla.suse.com/show_bug.cgi?id=1220996). The issue is a wicked boot-up time increase, when using `BOOTPROTO='dhcp+autoip'` from 15s to 30s, whenever wicked got a LEASE from the DHCPv4 server.


mtomaschewski (approved):
LGTM, thanks!

PR #1004, opened 2024-03-14
Replace "nvme show-nbft" with "nvme nbft show"

mwilck:
Upstream nvme-cli has moved the nbft support into a plugin.
The syntax is now "nvme nbft show" rather than "nvme show-nbft".

See https://bugzilla.suse.com/show_bug.cgi?id=1221358


mtomaschewski (approved):
LGTM, thanks!

PR #1005, opened 2024-03-18
Move all __attribute__ definitions to compiler.h

cfconrad:
Code cleanup/preparation for logging.h changes.


mtomaschewski (changes requested):
I'd try to consistently use a `NI_WICKED_` (we should cleanup other places in separate pull). In this case I'd use  `NI_WICKED_COMPILER_H` than just `NI_COMPILER_H`.

mtomaschewski (approved):
LGTM, thanks!

PR #1006, opened 2024-03-18
hide secrets in debug log (bsc#1221194) 

cfconrad:
Replaces cdata in a copy of a config node and it's children that contain passwords before logging it.

Please merge https://github.com/openSUSE/wicked/pull/1005 first.


mtomaschewski (commented):
[src/appconfig.c](https://github.com/openSUSE/wicked/pull/1006/files#diff-ce1c72e58a49310b57a3935e57d021014d365a8329aa657e3165dfdcad979508) is about the `/etc/wicked/*.xml` appconfig files -- extension definition  config in this case and not about interface configs/polices.

mtomaschewski (changes requested):
See previous comment. Commit  d1047fc340f162c3d2397a8bebd2173b705b7ffd is not needed and I'd not add it.
The xml does not contain any passwords (which would be valid for all users) and will never do. It's useless to **clone** the appconfig extension definition for nothing.

mtomaschewski (approved):
LGTM,  thanks!

PR #1007, opened 2024-04-03
client: fix to not convert sec to msec twice (bsc#1222105)

mtomaschewski:
Fix to not convert user-timeout in seconds into to timer timeout in millisecond twice.
It caused to wait 1000 times longer when a bridge with stp=on starts and all it's ports
are disconnected and unable to find carrier.


rma-x (approved):
Looks OK to me, although it is not always obvious without looking at a larger context which time units are used by a variable or function.

PR #1010, opened 2024-04-30
cleanup: fix enum-int-mismatch warnings

mtomaschewski:
- Fix inconsistent function declarations with unsinged int instead of ni_fsm_state_t
- Remove unused ni_fsm_state_t state argument artefact from ni_client_state_new
- Remove __ prefix from NI_FSM_STATE_MAX enum constant.


cfconrad (approved):
LGTM

PR #1016, opened 2024-05-22
 compat-suse: fix dummy type detection from ifname 

cfconrad:
* If the new `DUMMY=yes/no` or deprecated `INTERFACETYPE=dummy` variable aren't set, consider only `/^dummy[0-9]+$/` interface names as dummy and do not cause conflicts with e.g. correct vlan config on `dummy0.42` interfaces.
* Create `ifcfg-dummy.5.in` from markdown file.


mtomaschewski (dismissed):
LGTM, thanks!

mtomaschewski (approved):

PR #1017, opened 2024-05-27
client: warn if interface wait time reached

cfconrad:


mtomaschewski:
Pipeline is running to apply it to http://download.opensuse.org/repositories/network:/wicked:/testing/

mtomaschewski (approved):
OK, thanks!

PR #1018, opened 2024-05-29
man: (re)generate man pages from md sources

mtomaschewski:


Makefile.am
```diff
@@ -43,12 +43,20 @@ $(distdir)/ChangeLog: check-git-tool
 	  LANG=C git log -- $(top_srcdir) > $(distdir)/ChangeLog ; \
 	fi
 
+warn-pre-built-man:
```
cfconrad, line 46: I think this general warning is to much. As we build the manpages from `man/*.in` and not from `*.md`

man/Makefile.am
```diff
@@ -1,72 +1,93 @@
...
+templates_md_src		= $(addprefix src/,$(generated_md_man:=.md))
+templates_md_in			= $(generated_md_man:=.in)
+templates_in			= $(generated_in_man:=.in)
 
-templates			= $(generated:=.in)
+# pandoc would write @VAR@ as [at]VAR[at] to .in:
+#   .md sources are using %VAR% variables, while
+#   .in templates are using common autoconf @VAR@
+substitute_md_vars		= \
+	$(AM_V_GEN)sed		-e "s|[%]\(.*\)[%]|@\\1@|g"
```
cfconrad, line 66: The replacement is to bold. Take a look to https://github.com/openSUSE/wicked/blob/3585f97e533dcb7ee80d5096b145e1500b9d476e/man/src/ifcfg-wireless.5.md?plain=1#L242 which is in trouble with this regex.

So something like:
```suggestion
	$(AM_V_GEN)sed -E 's/%(wicked_(piddir|sbindir|statedir|storedir|schemadir|configdir|supplicantdir|extensionsdir)|PACKAGE_BUGREPORT)%/@\1@/g'
```
or
```suggestion
	$(AM_V_GEN)sed -E			\
	-e "s|%(wicked_piddir)%|@\1@|g"		\
	-e "s|%(wicked_sbindir)%|@\1@|g"	\
	-e "s|%(wicked_statedir)%|@\1@|g"	\
	-e "s|%(wicked_storedir)%|@\1@|g"	\
	-e "s|%(wicked_schemadir)%|@\1@|g"	\
	-e "s|%(wicked_configdir)%|@\1@|g"	\
	-e "s|%(wicked_supplicantdir)%|@\1@|g"	\
	-e "s|%(wicked_extensionsdir)%|@\1@|g"	\
	-e "s|%(PACKAGE_BUGREPORT)%|@\1@|g"
```

man/Makefile.am
```diff
@@ -1,72 +1,93 @@
...
-	@echo "substituting variables in $< > $@"
-	@rm -f $@ $@.tmp
-	@$(substitute_vars) < $< > $@.tmp && mv $@.tmp $@
+if pandoc
+%.in: src/%.md $(top_builddir)/config.status
+	@if $(AM_V_P); then echo "generating manual page: $< > $@" ; fi
+	@rm -f $@ $@.tmp $@.tmp
+	@$(AM_V_GEN)$(PANDOC) -s -f markdown -t man $< -o $@.tmp
+	@$(substitute_md_vars) -i "$@.tmp" && mv $@.tmp $@
+endif
```
cfconrad, line 91:
```suggestion
if pandoc
%.in: src/%.md
	@if $(AM_V_P); then echo "generating manual page: $< > $@" ; fi
	@rm -f $@
	@$(AM_V_GEN)$(PANDOC) -s -f markdown -t man $< -o $@
	@$(substitute_md_vars) -i "$@"

else
%.in: src/%.md
	$(error Missing pandoc to regenerate $@ from $< )
endif
```

I would simply error out here if there is no pandoc. Such developer should simply do not toucht the `*.md` files.
I think then this error from 3585f97e533dcb7ee80d5096b145e1500b9d476e is not needed anymore.

And I removed the dependency to `$(top_builddir)/config.status`, not sure why you had it there?

Makefile.am
mtomaschewski: No, we build from .md when pandoc is available. Only when pandoc is not available, we omit the .md -> .in step
and then warn when the dist archive is prepared for release with "make package" or "make rpmpackage" (with distcheck and other checks).

man/Makefile.am
mtomaschewski: > I would simply error out here if there is no pandoc.
No, the suggestion breaks **any build** -- on systems without pandoc and this are all versions < sle15-sp5 and even more: any RPM build, as wicked.spec does not have a BuildRequire enabled by default.

I do not want to convert man pages from md when we build an RPM using the released tar archive (and probably differently due to different pandoc version, ...) -- the .in files are BUILT_SOURCES included in the dist-tar archive created by "make dist" and we simply fill out the variables only while rpmbuild.

> Such developer should simply do not toucht the *.md files.
With the suggestion you don't need to touch the .md files at all. Just try out: `./autogen.sh --disable-pandoc && make` -> error. Same will happen when pandoc is not available (in the build chroot / container, ...) in the system.

> And I removed the dependency to $(top_builddir)/config.status, not sure why you had it there?
This is IMO needed to automatically rebuild when make decides to make a dependency configure re-run while pandoc is available (and we want to rebuild from md). It's required by "distcheck" as the .in files are built sources.

mtomaschewski:
```
# ./autogen.sh --disable-pandoc && make distcheck
[...]
make[2]: Entering directory '/root/wicked/git/man'
  GEN      ifcfg-bond.5
  GEN      ifcfg-bonding.5
  GEN      ifcfg-bridge.5
  GEN      ifcfg-dhcp.5
  GEN      ifcfg-infiniband.5
  GEN      ifcfg-lo.5
  GEN      ifcfg-macvlan.5
  GEN      ifcfg-macvtap.5
  GEN      ifcfg-ovs-bridge.5
  GEN      ifcfg-ppp.5
  GEN      ifcfg-tunnel.5
  GEN      ifcfg-vlan.5
  GEN      ifcfg-vxlan.5
  GEN      ifcfg.5
  GEN      ifdown.8
  GEN      ifroute.5
  GEN      ifrule.5
  GEN      ifstatus.8
  GEN      ifsysctl.5
  GEN      ifup.8
  GEN      routes.5
  GEN      wicked-config.5
  GEN      wicked-ethtool.8
  GEN      wicked.7
  GEN      wicked.8
  GEN      wickedd.8
Makefile:725: *** Missing pandoc to regenerate ifcfg-dummy.5.in from src/ifcfg-dummy.5.md .  Schluss.
make[2]: Leaving directory '/root/wicked/git/man'
Makefile:592: recipe for target 'distdir' failed
make[1]: *** [distdir] Error 1
make[1]: Leaving directory '/root/wicked/git'
Makefile:687: recipe for target 'dist' failed
make: *** [dist] Error 2
```

Makefile.am
cfconrad: Ok I had different expectation. But I guess it makes sense to (re-)build the manpages with pandoc, if we do not explicit disable it with `./configure --disable-pandoc`

cfconrad (approved):
LGTM

PR #1020, opened 2024-06-05
arp:  increase arp-send retry value (bsc#1218668) 

cfconrad:
If the kernel reports `ENOBUFS` while sending the arp package, we do not want to treat this as a fatal error. But we also don't want to accept `ENOBUFS` forever. 

The previous retry value of `3` wasn't sufficient, as the kernel reports "up & running" for an interface with underlying
bonding in 802.3ad mode to early ([bsc#1218668](https://bugzilla.suse.com/show_bug.cgi?id=1218668)) and returns `ENOBUFS` in attempts to send a package over this link.

With the new default value of `10`, we give the driver a minimum amount of at least 6s to accept packages.


mtomaschewski (approved):
Increasing notify retries should not be needed as it's `verify && apply && notify` flow, but OK, should not hurt.

Merged to testing, RPMs will appear at http://download.opensuse.org/repositories/network:/wicked:/testing/

Waiting for customer feedback.

PR #1021, opened 2024-06-12
man: add supported bonding options to ifcfg-bonding(5) man page

cfconrad:
- generate manpage from *.md
- add list of supported bonding options (thx @mtomaschewski for co-authoring)


mtomaschewski (changes requested):
We also need to adopt `man/Makefile.am` and move the `ifcfg-bonding.5` manpage from `generated_in_man` into `generated_md_man` list.

mtomaschewski (approved):
LGTM, thanks!

PR #1022, opened 2024-06-17
arp: increase arp-send retry value (bsc#1218668) 

cfconrad:
* arp: increase  to 16 retries result in at least 10s
* capture: log individual send errors from kernel as debug

This is a follow-up to https://github.com/openSUSE/wicked/pull/1020


mtomaschewski (approved):
LGTM, thanks!

PR #1025, opened 2024-07-15
ipv6: fix setup on ipv6.disable=1 kernel cmdline (bsc#1225976) 

mtomaschewski:


cfconrad (approved):
LGTM and verified with tw

PR #1026, opened 2024-07-16
client: fix to set and use origin in loaded config (bsc#1226125)

mtomaschewski:
When loading and migrating a master xml configuration with
obsolete port array and missed port interface configurations,
the migration hook were unable to inherit the origin from
master configuration cleared by a "read raw config" request,
failed to apply the port configurations caused a no-carrier
setup state on the (bond) master.


cfconrad (approved):
LGTM, thx!

PR #1029, opened 2024-07-19
wicked.spec.in: update date to 2024

cfconrad:
Last commit in `wicked.spec.in` was on May 2024, I guess we simply forgot to update the date.


mtomaschewski (approved):
OK

PR #1032, opened 2024-08-27
systemd: ensure wicked launches after dbus service

simotek:
After the change to dbus-broker it was possible for wicked to start before the dbus-broker service was running. Adding `BindsTo: dbus.socket` fixes this.

https://bugzilla.suse.com/show_bug.cgi?id=1229745
https://bugzilla.suse.com/show_bug.cgi?id=1229799
https://bugzilla.suse.com/show_bug.cgi?id=1229800

### DIFF
--- a/etc/systemd/wickedd-auto4.service.in
+++ b/etc/systemd/wickedd-auto4.service.in
@@ -1,10 +1,9 @@
 [Unit]
 Description=wicked AutoIPv4 supplicant service
-Requisite=dbus.service
+BindTo=dbus.service
 After=local-fs.target dbus.service network-pre.target SuSEfirewall2_init.service
 Before=wickedd.service wicked.service network.target
 PartOf=wickedd.service
-BindsTo=dbus.service
 
 [Service]
 Type=notify
--- a/etc/systemd/wickedd-auto4.service.in
+++ b/etc/systemd/wickedd-auto4.service.in
@@ -4,6 +4,7 @@ Requisite=dbus.service
 After=local-fs.target dbus.service network-pre.target SuSEfirewall2_init.service
 Before=wickedd.service wicked.service network.target
 PartOf=wickedd.service
+BindsTo=dbus.service
 
 [Service]
 Type=notify
--- a/etc/systemd/wickedd-dhcp4.service.in
+++ b/etc/systemd/wickedd-dhcp4.service.in
@@ -1,10 +1,9 @@
 [Unit]
 Description=wicked DHCPv4 supplicant service
-Requisite=dbus.service
+BindTo=dbus.service
 After=local-fs.target dbus.service network-pre.target SuSEfirewall2_init.service
 Before=wickedd.service wicked.service network.target
 PartOf=wickedd.service
-BindsTo=dbus.service
 
 [Service]
 Type=notify
--- a/etc/systemd/wickedd-dhcp4.service.in
+++ b/etc/systemd/wickedd-dhcp4.service.in
@@ -4,6 +4,7 @@ Requisite=dbus.service
 After=local-fs.target dbus.service network-pre.target SuSEfirewall2_init.service
 Before=wickedd.service wicked.service network.target
 PartOf=wickedd.service
+BindsTo=dbus.service
 
 [Service]
 Type=notify
--- a/etc/systemd/wickedd-dhcp6.service.in
+++ b/etc/systemd/wickedd-dhcp6.service.in
@@ -1,10 +1,9 @@
 [Unit]
 Description=wicked DHCPv6 supplicant service
-Requisite=dbus.service
+BindTo=dbus.service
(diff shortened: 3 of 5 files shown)


cfconrad:
Thx for the PR. I think we should do the same for `wickedd-auto4.service.in`, `wickedd-dhcp4.service.in`, `wickedd-dhcp6.service.in`, `wickedd-nanny.service.in` .

Also, I don't see any need of `Requisite=dbus.service` (introduced with 71d63181220425e8acb5e9439c6b6e4adc37b818) in these files anymore. As from what I understand, [`BindTo=`](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#BindsTo=) also lead to a restart of wickedd, if the dbus services might get restarted.

But I'm missing, why is dbus-broker different? Does it report "start-up success" much earlier and being "active" so much later?

simotek:
Added the other spec files as well.

drankinatty:
Is there a problem with `BindsTo=` for those who have network configured to **start at boot** irrespective of user login?

My reading is that if we do `BindsTo=dbus.service` in **wicked.service** than any time **dbus.service** is stopped or crashes, then **wicked.service** is also stopped automatically. How would that affect wicked's control of wireless devices started at boot?

What about Andrei's suggestion of swapping `dbus.socket` for `dbus.service` in the **wickedd....** files? Would that be better for interfaces configured to start `on-boot`?

I don't have the expertise to say one would or wouldn't but just thinking though how `BindsTo=` makes wicked depended on dbus being up and how a dbus restart would also trigger a wicked restart seemed a bit odd?

Just food for thought. You all know better than I.

simotek:
> Is there a problem with `BindsTo=` for those who have network configured to **start at boot** irrespective of user login?
> 
> My reading is that if we do `BindsTo=dbus.service` in **wicked.service** than any time **dbus.service** is stopped or crashes, then **wicked.service** is also stopped automatically. (meaning a restart of dbus would also trigger a complete restart of wicked?) How would that affect wicked's control of wireless devices started at boot?

dbus-broker is configured to refuse manual start stop, far to much of user space can't cleanly  handle dbus restarting for it to be something we'd currently consider safe.

We are also only talking about a limited number of Tumbleweed systems here, the dbus-broker switch won't happen in SLE / Leap 15 and currently I don't believe we are planning to include wicked in SLE-16 

> What about Andrei's suggestion of swapping `dbus.service` with `dbus.socket` in the **wickedd....** files? Would that be better for interfaces configured to start `on-boot`?
> 
> I don't have the expertise to say one would or wouldn't but just thinking though how `BindsTo=` makes wicked depended on dbus being up and how a dbus restart would also trigger a wicked restart seemed a bit odd?
> 
> Just food for thought. You all know better than I.

If the wicked maintainers are happy with this, then it is certainly the way most other services handle this.

mtomaschewski:
> Is there a problem with `BindsTo=` for those who have network configured to **start at boot** irrespective of user login?
> 
/
> My reading is that if we do `BindsTo=dbus.service` in **wicked.service** than any time **dbus.service** is stopped
> or crashes, then **wicked.service** is also stopped automatically. (meaning a restart of dbus would also trigger a complete restart of wicked?)

You mix **wicked.service** (aka network.service) and **wickedd.service** (backend daemons used by wicked service).

> How would that affect wicked's control of wireless devices started at boot?

No. Stopping `wickedd*.service` does not stop network, but the backend daemons only. Stopping them, does not
stop any interfaces, but the daemons only: yes, the daemons will not e.g. renew leases when stopped.

mtomaschewski:
> > Thx for the PR. I think we should do the same for `wickedd-auto4.service.in`, `wickedd-dhcp4.service.in`, `wickedd-dhcp6.service.in`, `wickedd-nanny.service.in` .
> 
> `wickedd-nanny.service.in` can't start before `wickedd.service.in` so its probably ok, but i'll add it to the others shortly.

This IMO does not matter. All `wickedd*` services **have** dependency to dbus via `Requisite=dbus.service`, a variant similar to Requires, but not triggering dbus start, but failing when it's not running. A `BindsTo=` is also such a variant,
same nature, just a bit different "detail behavior" again.

Means, I'd tend to consistently change all services with: `sed -e "s|Requisite=dbus.service|BindsTo=dbus.service|g"`
as already mentioned by Clemens in https://github.com/openSUSE/wicked/pull/1032#issuecomment-2312388618

mtomaschewski (changes requested):
Please consistently change the current `Requisite` to the `BindsTo` dependency:
```
sed -e "s|Requisite=dbus.service|BindsTo=dbus.service|g" -i wicked*.service
```

cfconrad (approved):

mtomaschewski (approved):
LGTM now, thanks!

mtomaschewski:
> LGTM now, thanks!

Ahm... I didn't noticed that it's using `BindTo=` instead of correct `BindsTo=`... :-/

PR #1036, opened 2024-09-20
arp: don't set target broadcast hardware address

mtomaschewski:
According to RFC5227 section 2.1.1, the "'target hardware address' field is ignored and SHOULD be set to all zeroes".
The section 2.3 defines that an announcement is identical, except of the announced IP in sender and target IP field.


cfconrad (approved):
Good catch

PR #1038, opened 2024-10-07
dhcp4: add ignore-rfc3927-1-6 wicked-config(5) option (jsc#PED-10855)

mtomaschewski:
The RFC3927 Section 1.6: Alternate Use Prohibition prohibits the use of link-local
IPv4 (zeroconf) address range for other purposes, such as for DHCP.
Enabling this option in wicked-config(5) permits to use the addresses in DHCP.

### DIFF
--- a/man/wicked-config.5.in
+++ b/man/wicked-config.5.in
@@ -392,6 +392,19 @@ lease time to one hour:
 .PP
 
 .TP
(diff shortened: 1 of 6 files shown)


cfconrad (approved):
LGTM, thx

PR #1040, opened 2024-10-08
Sysctl code cleanup - remove unused code

cfconrad:
Removing unused code to read sysctl's from procfs. This code wasn't used since 8 years (see fe0be3eda9c76deacda9d66ac117435017df9cef)


mtomaschewski (approved):
Yes, wicked is receiving the sysctl's via rtnetlink, the active reading of each sysctl were needed on attic (SLE-11) kernels only and the code apparently has been "disarmed" by inserting a `return 0` "in the middle" during testing, but without to remove it in final commits.

PR #1043, opened 2024-10-25
compat-suse: use iftype in sysctl handling (bsc#1230911)

cfconrad:
Fix loopback sysctl configuration. Since gh#openSUSE/open-build-service#895 sysctl settings 
for loopback device where ignored. This is considered wrong as it fails to set `disable_ipv6` properly.
On top we doesn't set the ipv6/ipv4 enabled node since gh#openSUSE/open-build-service#990 causing
lo to stay in `device-not-running`, with `net.ipv6.conf.(all|default).disable_ipv6=1`.

Fixes:
* Always generate the ipv4/ipv6 `<enabled>true|false</enabled>` node
* Inherit all,default,interface sysctl settings also for loopback,
  except for `use_tempaddr` and `accept_dad` sysctl's.
* Consider only interface specific `accept_redirects` sysctl settings.
* Adopt ifsysctl(5) manual page with wicked specific behavior.


mtomaschewski (approved):
LGTM, we've already extensively tested and reviewed it before.

PR #1046, opened 2024-11-15
route: use new array macros and fix on error leaks

mtomaschewski:
- Avoid unsafe use of append(array, ref(entry)) function calls not allowing to release reference
  on append failure in favor of the new append_ref(array, obj) and insert_ref macros.
- Free discarded duplicate route and route rule option array and on parse errors.
- Remove table list argument from route create for consistent reference counting behavior.


cfconrad (approved):
Nice cleanup, thx.

PR #1047, opened 2024-12-03
fsm: use refcount and array macros in worker and policy

mtomaschewski:
Refactored and cleaned up to use our common refcounting and array macros
so the worker and policy are using the same function prototypes, e.g. same
names and meaning of array remove vs. delete.


cfconrad (approved):
LGTM

PR #1048, opened 2024-12-16
ifxml: migrate and generate lower configs/policies

mtomaschewski:
- migrate master and lower references in policy matches
- generate missing config/policy for lower interfaces
- warn when missing config has been generated


cfconrad (approved):
LGTM

PR #1049, opened 2024-12-16
fsm: remove children reference array from worker 

mtomaschewski:
The fsm worker is a structure that contains config and/or device.
Use the master/lowerdev worker config interface references only
and remove the confusing children array causing complex cleanup.


cfconrad (approved):
LGTM

PR #1059, opened 2025-02-12
xml: adjust to refcount macros in xml_node

mtomaschewski:


cfconrad (approved):
Thanks, 
for transparency, the function `xml_node_new()` is using `xmalloc()` which is still on our TODO to cleanup

PR #1060, opened 2025-02-25
policy: cleanup fsm-policy match and transform processing

mtomaschewski:
- Fix crash and leak on error in policy action processing when transforming
  them to effective config (not triggered by policies generated from ifcfg).
- Cleanup applicable fsm policy matching to use fsm_policy_array in recheck
- Adjust fsm function names acting on policies vs fsm-policy functions.
- Changed to replace the inner fsm-policy in managed policy update instead
  to tweak it's content, to not require content changed compare later.


cfconrad (approved):
Thx, looks good to me!

PR #1063, opened 2025-09-04
ovs: suppress harmless ovs-vsctl discover call errors

cfconrad:
This commit changes error messages from failed `ovs-vsctl`
discovery calls to debug logs. These temporary failures occur
when `wicked` queries `ovs` for ports that no longer exist,
causing false CI failures without affecting service stability.

### DIFF
--- a/src/ovs.c
+++ b/src/ovs.c
@@ -202,7 +202,7 @@ ni_ovs_vsctl_bridge_to_vlan(const char *brname, uint16_t *vlan)
 	rv = ni_process_run_and_capture_output(pi, &buf);
 	ni_process_free(pi);
 	if (rv) {
-		ni_error("%s: unable to query bridge vlan", brname);
+		ni_debug_application("%s: unable to query bridge vlan", brname);
 		goto failure;
 	}
 
@@ -264,7 +264,7 @@ ni_ovs_vsctl_bridge_to_parent(const char *brname, char **parent)
 	rv = ni_process_run_and_capture_output(pi, &buf);
 	ni_process_free(pi);
 	if (rv) {
-		ni_error("%s: unable to query bridge parent", brname);
+		ni_debug_application("%s: unable to query bridge parent", brname);
 		goto failure;
 	}
 
@@ -320,7 +320,7 @@ ni_ovs_vsctl_bridge_ports(const char *brname, ni_netdev_ref_array_t *ports)
 	rv = ni_process_run_and_capture_output(pi, &buf);
 	ni_process_free(pi);
 	if (rv) {
-		ni_error("%s: unable to query bridge ports", brname);
+		ni_debug_application("%s: unable to query bridge ports", brname);
 		goto failure;
 	}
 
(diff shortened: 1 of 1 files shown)


mtomaschewski:
>  due to initialization race conditions

This are IMO not **_initialization_** race conditions.

It's about the fact, that we execute `ovs-vsctl` commands and process multiple (NEWLINK)  rtnetlink kernel events
that it triggers, where the event data may still contain relations that do no exist in ovs any more and attempts to
resolve the relation to the bridge cause an ovs-vsctl query to return a failure and trigger such an error.

The most significant error reported here in the CI may happen due to a port removal from bridge:

When a port (ens5) is added to a bridge (ovsbrA):
```
Sep 08 22:16:38.413449 install ovs-vsctl[4432]: ovs|00001|vsctl|INFO|Called as /usr/bin/ovs-vsctl --may-exist add-port ovsbrA ens5
```
in the kernel, the port master property does not point to the bridge, but to the `ovs-system` datapath and the bridge relation is visible in ovs (via ovs-vsctl query) only:
```
3: ens5: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast master ovs-system state UP group default qlen 1000
```

When we remove the port from the bridge:
```
Sep 08 22:16:38.644943 install ovs-vsctl[4484]: ovs|00001|vsctl|INFO|Called as /usr/bin/ovs-vsctl del-port ovsbrA ens5
Sep 08 22:16:38.650537 install wickedd[783]: subprocess 4484 (/usr/bin/ovs-vsctl del-port ovsbrA ens5) exited with status 0 [0m0.008s]
```
there may be still an (port interface update NEWLINK, e.g. carrier state) rtnetlink event for the port, that still contains
the relation to ovs-system, which can't be resolved in ovs to a bridge any more:
```
Sep 08 22:16:38.661252 install ovs-vsctl[4485]: ovs|00001|db_ctl_base|ERR|no port named ens5
Sep 08 22:16:38.661695 install wickedd[783]: subprocess 4485 (/usr/bin/ovs-vsctl port-to-br ens5) exited with status 1 [0m0.010s]
Sep 08 22:16:38.661701 install wickedd[783]: ens5: unable to query port bridge
```
causing to report the `unable to query port bridge` error in wicked due to exit status 1 reported by ovs-vsctl.

This `ens5` to `ovs-system` master relation gets removed in a further rtnetlink (NEWLINK) event.

mtomaschewski (commented):
We can apply it, but I wanted to wait until you're back to discuss it.

Does it really make sense to lower it to warning for the ovs **queries** just to make CI grepping for all wicked error messages happy? There is IMO no reason to log them at all.

Why not lower it to debug / let the caller decided if the negative reply from ovs is worth a **2nd** error message from wicked -- **1st** one is already logged by `ovs-vsctl` itself. They also happen only when wickedd received some event from kernel with a reference, which do not exist any more in ovs in the meantime ... "since 0.01s".
I were unable to trigger them by executing the testcases on my machines... you?

cfconrad:
I'm fine with debug as well. 
No I wasn't able to trigger it on my personal test-machines as well. I only see it (in regular bases) in the CI.

mtomaschewski:
Even we'd remove them completely, there is still:
`Sep 08 22:16:38.661252 install ovs-vsctl[4485]: ovs|00001|db_ctl_base|ERR|no port named ens5`

and a `exited with status 1` debug message visible in the logs:
`Sep 08 22:16:38.661695 install wickedd[783]: subprocess 4485 (/usr/bin/ovs-vsctl port-to-br ens5) exited with status 1 [0m0.010s]`

cfconrad:
Ok, I will change the log message to debug.

mtomaschewski (approved):
LGTM now, thanks!

PR #1064, opened 2025-09-05
client: add iaid show,get --iaid-format option

mtomaschewski:
- Add the `--iaid-format <dec,hex,mac>` option in show and get actions
  to format the IAID value as decimal (default), a `0x<hex>` hex number
  or as a mac-like hex-string (`xx:xx:xx:xx`).
- Add parsing of the mac-like hex-string (`xx:xx:xx:xx`) in set action.

### DIFF
--- a/client/iaid.c
+++ b/client/iaid.c
@@ -1,7 +1,7 @@
 /*
  *	wicked client iaid commands
  *
- *	Copyright (C) 2017 SUSE LINUX GmbH, Nuernberg, Germany.
+ *	Copyright (C) 2017-2025 SUSE LLC
  *
  *	This program is free software; you can redistribute it and/or modify
  *	it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


client/iaid.c
```diff
@@ -92,19 +151,28 @@ ni_do_iaid_dump(int argc, char **argv)
...
 	};
 	int opt = 0, status = NI_WICKED_RC_USAGE;
 	ni_iaid_map_t *map = NULL;
 	const char *ifname = NULL;
 	unsigned int iaid;
+	ni_iaid_fmt_t opt_iaid_fmt = OPT_IAID_FMT_DEC;
 
 	optind = 1;
-	while ((opt = getopt_long(argc, argv, "+h", options, NULL)) != EOF) {
+	while ((opt = getopt_long(argc, argv, "+hx", options, NULL)) != EOF) {
```
cfconrad:
```suggestion
	while ((opt = getopt_long(argc, argv, "+h", options, NULL)) != EOF) {
```
I don't see the reason for this `x`?

mtomaschewski: It was an leftover from another variant.

cfconrad (approved):
LGTM

PR #1068, opened 2026-02-03
netlink: remove libnl-route dependency

mtomaschewski:
Remove the (only/remaining) `libnl-route` function call of `nl_rtgen_request` to avoid automatic cache (de-)registrations at start and exit of all binaries.


cfconrad (approved):
Thx, LGTM

PR #1071, opened 2026-06-11
version 0.6.79

mtomaschewski:
- Fix an indirect remote shell command injection via unsanitized
  dhcp strings and leaseinfo dump ([bsc#1265221](https://bugzilla.suse.com/show_bug.cgi?id=1265221),[CVE-2026-44932](https://www.suse.com/security/cve/CVE-2026-44932.html)):
  - Fix to escape single-quotes in leaseinfo dump output used by the
    `wicked test dhcp4` and `wicked test dhcp6` and written to the
    /run/wicked/leaseinfo.* files, e.g. to pass them to netconfig.
    A netconfig modify filtered for strict key='value' lines without
    any escaped quotes and discarded these lines already before.
  - Fix posix-tz-dbname and tz-string option processing checks to
    permit only valid characters according to RFC4833.
  - Discard string values containing single-quotes in other options.
  - Trigger to regenerate initrd that may contain wicked binaries on
    updates from wicked versions <= 0.6.78.


cfconrad (approved):
LGTM, thx!

PR #1072, opened 2026-06-30
xpath: fix gcc-15.x assignment discards 'const' warning

mtomaschewski:
The `ident` variable used in `if ((colons = strstr(ident, "::")) != NULL)`
were declared as `const char *` pointing to a `char identbuf[…]`.

### DIFF
--- a/src/xpath.c
+++ b/src/xpath.c
@@ -82,7 +82,7 @@ static xpath_result_t *	__xpath_build_boolean(int);
 
 static xpath_enode_t *	__xpath_build_expr(const char **, char, int infixprio);
 static int		__xpath_enode_assert_element(xpath_enode_t **);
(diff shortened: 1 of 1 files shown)


src/xpath.c
```diff
@@ -463,7 +465,7 @@ __xpath_enode_assert_element(xpath_enode_t **epp)
 }
 
 
-const char *
+char *
 __xpath_next_identifier(const char **pp)
```
cfconrad, line 469: I'm not sure if it is a good idea to retrieve the static buffer as not `const`

cfconrad: I double checked your change. I think I get the idea of minimal change in here.

cfconrad (approved):

PR #1075, opened 2026-07-14
Conditionally install configs and scripts in /usr for Immutable Mode

mtomaschewski:
- Conditionally move extensions and hook scripts from /etc to 
  /usr/lib[exec]/wicked and create required /var and /etc 
  directories using systemd tmpfiles.d on openSUSE >= 16.0.
- Move /etc/wicked/{common,client,server,nanny}.xml wicked-config(5)
  files to %_distconfdir (/usr/etc/wicked) if it is defined.
- Enable dual-path lookup for wicked-config(5), extensions and hook 
  scripts in /etc/wicked and then their /usr paths when split 
  installation paths are configured.
- Add wicked-config(5) circular include, depth limit parser guards
  and support for directory-based includes `<include dir="..."/>`.


cfconrad (approved):
LGTM, thx.

-> updated man pages will follow

PR #1076, opened 2026-07-22
Document the Immutable Mode /usr config split in the manual pages

mtomaschewski:
- Document the conditional `/usr` vs `/etc` config, extensions and hook
  script split with dual-path search in `wicked-config(5)`, `ifcfg(5)`
  and `wicked-firmware(8)`.
- Add a `util/templatepp.gawk` preprocessor for variables and conditional
  blocks and use it instead of `sed` to render templates for both layouts.


util/templatepp.gawk
janvhs, line 1: The pre-processor doesn't catch unclosed if statements. Maybe add this with a patch like this https://github.com/openSUSE/wicked/commit/b9f88e3496f403feecce4c122451ad2387f8a9a9

janvhs (changes requested):
LGTM just a small improvement for the templating pre-processor

janvhs (approved):
LGTM

util/templatepp.gawk
mtomaschewski: picked ff.

PR #1077, opened 2026-07-22
version 0.6.80

mtomaschewski:
- Immutable Mode: conditionally install configs, scripts and extensions in /usr,
  document the /usr vs /etc config split in the manual pages, and render via new
  util/templatepp.gawk preprocessor
- xpath: fix gcc-15.x assignment discards 'const' warning
- nanny: use opaque id for managed-netif dbus path
- netlink: remove libnl-route dependency
- spec: guard the sle_version macro
- service: drop rcwicked* links on 16.x (jsc#PED-266)
- client: add iaid show,get --iaid-format option
- ovs: suppress harmless ovs-vsctl discover call errors
- teamd: fix TEAM_LACP_ACTIVE default to true
- policy: cleanup managed objectmodel, fsm-policy match and transform processing


janvhs (approved):
LGTM

PR #1080, opened 2026-09-17
dhcp4: resume fsm on link-up while in down state (bsc#1265576)

mtomaschewski:
When the link goes down while the dhcp4 supplicant confirms or reapplies
a lease, its fsm parks in DOWN state and waits for nanny to reissue an
acquire request on link-up. In rare cases nanny didn't, so the interface
stayed without a working lease.

On link-up, grant nanny a short grace period, then continue to confirm
or remove an expired lease from the interface and discover a new one.

Remove unusable lease / IP from interface also on NAK and ARP conflict
before acquiring a new one.


cfconrad (approved):
LGTM, already [tested](https://bugzilla.suse.com/show_bug.cgi?id=1265576#c19).

